PR #12, opened 2023-03-09
Update User_Experience.md

jolie-zh:


sophiaabolore (commented):
I think this was a really great description of the user experience domain. 

I think some ways it could be made even better is by encompassing the concepts of UX, Accessibility, and demographics as components in the broader designing process. Although there are differing views on what each stage of the design process is, one of the most popular is to 1. Empathize (understand the user and their needs) 2. Define (State Your Users' Needs and Problems - this is likely where the target demographic selection occurs) 3. Ideate—Challenge Assumptions and Create Ideas 4.  Prototype—Start to Create Solutions 5.  Test—Try Your Solutions Out (this is likely where testing for user experience and accessibility would come in). 

Here's a resource that might provide more information: https://www.interaction-design.org/literature/topics/design-thinking

PR #15, opened 2023-03-10
Provided information about agile framework and methodology

oishi37:
I added two resources for learning more about agile framework and methodology and provided brief description for each.


oishi37:
Should I remove them from software engineering and move them under development process? Also I am not sure what you mean by "Also I'd expect a cleanup of things like "potential topics" when you are committing to the file.". Further explanation on this would be really helpful.

PR #16, opened 2023-03-10
Adding Django.

mchan7841:


Topics/Tech_Stacks.md
```diff
@@ -1 +1,184 @@
 ## Tech Stacks
+
+## Django
```
jorjani, line 3: Please move this to a Django specific file (maybe even under Python) and use references and links freely instead of bringing the content from somewhere else into the repo. What matters is making it accessible for students in the future (and updated) rather than "writing" the content into this repo

jorjani (changes requested):

Topics/Tech_Stacks.md
mchan7841: As you suggested, this has been moved to a Django file and the use of links and references is more widespread for future use of the content.

PR #17, opened 2023-03-12
added salesforce api to techstack.md

CorinneLS:


Topics/Tech_Stacks.md
```diff
@@ -1 +1,47 @@
 ## Tech Stacks
+
+---
+
+# Access to Salesforce API
```
jorjani, line 5: Please create a Salesforce specific file and move these there! If you are borrowing from a page, you can use a link to the page rather than repeating the content here

jorjani (changes requested):

Topics/Tech_Stacks.md
CorinneLS: Would the salesforce specific file in the Topics folder?

PR #44, opened 2023-03-20
Building RESTful APIs using Django Rest Framework

sarihammad:
A small introduction on the process of building RESTful APIs using Django Rest Framework. Includes official documentation and helpful tutorials.

### DIFF
--- a/Topics/Tech_Stacks/Django_Rest.md
+++ b/Topics/Tech_Stacks/Django_Rest.md
@@ -12,7 +12,9 @@ Django Rest Framework is a popular library for building RESTful APIs with Django
 
 ### Set-Up
 
- We first set up the virtual environment and activate it using ```python3 -m venv myenv ``` and ```source myenv/bin/activate``` respectively in the project directory.
+ We first Install virtualenv by running ```python3 -m pip install virtualenv```. This will ensure that we have the necessary package installed to create and manage virtual environments.
+
+ We now set up the virtual environment and activate it using ```python3 -m virtualenv –p `which python3.11` venv ``` and ```source venv/bin/activate``` respectively in the project directory. Note that ```-p `which python3.11` ``` is placed to avoid common issues with python file directory locations. Also, you can replace `python3.11` with the latest available version of python.
 
  We now install Django and Django Rest Framework using the command ```pip install django djangorestframework```.
 
@@ -60,7 +62,7 @@ class TodoItem(models.Model):
     completed = models.BooleanField(default=False)
 ```
 
-Next, we'll create a serializer to convert our model instances to JSON. We now create a new file called serializers.py in the app API folder and add the following code:
+Assuming we want to create an API for your TodoItem model that allows clients to create, read, update, and delete TodoItem instances. We may define a serializer class in serializers.py in the app API folder that converts TodoItem instances to JSON. We add the following code in serializers.py:
 
 ``` {python}
 from rest_framework import serializers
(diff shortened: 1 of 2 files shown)


Topics/Tech_Stacks/Django_Rest.md
```diff
@@ -0,0 +1,120 @@
...
+
+ - [Python](https://www.python.org/): High level general purpose programming language.
+ - [Django](https://www.djangoproject.com/): Web framework for Python.
+
+### What Is Django Rest Framework?
+Django Rest Framework is a popular library for building RESTful APIs with Django. It provides powerful tools for building APIs, including serialization, authentication, and permissions. You can easily create APIs that support multiple formats like JSON and XML that can be consumed by a variety of client applications. To demonstrate the power of Django Rest Framework, we will create a simple API.
+
+### Set-Up
+
+ We first set up the virtual environment and activate it using ```python3 -m venv myenv ``` and ```source myenv/bin/activate``` respectively in the project directory.
```
nzhou27, line 15: Please either add a little bit more info about the constraints on setting up virtual env. When I was studying setting up virtual env with python, I had a problem where my python file isn't locating at the same directory as others and some of my peers had the same issue, but most tutorial doesn't mention this. It is critical to check out the directory for where the python file is located and make sure your virtual env points to the right location and the right version too. Please check out this resource for reference: https://help.dreamhost.com/hc/en-us/articles/115000695551-Installing-and-using-virtualenv-with-Python-3    (mainly from step 1 to 4)

Topics/Tech_Stacks/Django_Rest.md
```diff
@@ -0,0 +1,120 @@
...
+Before learning Django Rest Framework, make sure that you have a working knowledge of the following:
+
+ - [Python](https://www.python.org/): High level general purpose programming language.
+ - [Django](https://www.djangoproject.com/): Web framework for Python.
+
+### What Is Django Rest Framework?
+Django Rest Framework is a popular library for building RESTful APIs with Django. It provides powerful tools for building APIs, including serialization, authentication, and permissions. You can easily create APIs that support multiple formats like JSON and XML that can be consumed by a variety of client applications. To demonstrate the power of Django Rest Framework, we will create a simple API.
+
+### Set-Up
+
```
nzhou27, line 14: need to pip install virtualenv as well

Topics/Tech_Stacks/Django_Rest.md
```diff
@@ -0,0 +1,120 @@
...
+from .models import TodoItem
+
+class TodoItemSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = TodoItem
+        fields = ('id', 'title', 'completed')
+```
+
+Next, we'll create a view to handle incoming HTTP requests (GET, POST, PUT, PATCH and DELETE). We create a new file called views.py in the API app folder and add the following code:
+
```
nzhou27, line 76: it would be very helpful if you could provide a small but complete example in how to use serialization to organize your code as it is a very important part of Django, maybe even with a small customized query selection to show that  we can easily filter out the data we want from database. I see you importing the serializer here but actually providing a small chunk of sample code would be very helpful

nzhou27 (changes requested):
Hello, 
I think you made a very nice and yet detailed post about Django and usage of virtual env, but for the serialization part, using the import in views.py but not mentioning what it does is a little bit confusing for newbies, and there are some other minor detail with setting up virtualenv. 

Please kindly make some changes and let me know if you have any questions.

 Thank you

Topics/Tech_Stacks/Django_Rest.md
sarihammad: I also ran into this issue before. I added more details on setting up the virtual environment.

Topics/Tech_Stacks/Django_Rest.md
sarihammad: I added a more detailed explanation on serialization, and gave an example with a customized query selection.

PR #54, opened 2023-03-20
Add DynamoDB for Dummies

spalmurray:
A simple overview of DynamoDB along with some links to DynamoDB resources I've found helpful.


yuengeoff (changes requested):
Overall, this is well-written and provides useful information and helpful resources for readers to learn about DynamoDB. With some slight changes regarding the title, structure, and some additional examples, this could be an even more valuable resource for those looking to use DynamoDB effectively

PR #64, opened 2023-03-21
Added User Story Section to Software_Engineering.md

jonathanginevro:
Overview of user stories and the role that they play in product management. This includes descriptions of each of the major sections of a user story and insight into how they are created and maintained. Additionally, examples and templates are included throughout the document to demonstrate how to write effective user stories.


Topics/Product_Management.md
```diff
@@ -1 +1,73 @@
 ## Resources for Product Management
```
SanjanaGirish, line 1: You can remove the starting line stating 'Resources for product management'

SanjanaGirish (changes requested):
Great resources and the topic is well structured. Please resolve conflicts with the main branch. A helpful suggestion would be to add the topic in a separate file and link it to the main page to avoid conflicts. 

Also, I would like to hear more about why User stories are added to the Product Management section. Is it because product managers typically write user stories? There is no link to product management drawn in the topic. I think User stories fit in the 'Development processes' category better but would like to hear your thoughts.

jonathanginevro:
Thank you for the feedback. Yes, as you mentioned, user stories would typically be created by the product manager (or product owner), and so I thought it fit best in this document. Also, my original idea to work on this topic came from the product management lecture slides, which included a subtopic on user stories. 

However, seeing what has now been merged into the Product Management document, I agree that this topic would be out of place. I hadn't considered the 'Development Processes' document but think that would work well. I also think that my section would work well if it were linked to the 'Software Engineering' document. It would probably fit best under the Scrum category since user stories are generally associated with that framework.

jonathanginevro:
Changed file location and resolved merge conflicts

SanjanaGirish:
Great job, all feedback has been incorporated well.

PR #76, opened 2023-03-22
Added Axios.md

BKPrograms:
Discussed Axios and it's uses, along with providing code examples and resource links

### DIFF
--- a/Topics/Tech_Stacks/Axios.md
+++ b/Topics/Tech_Stacks/Axios.md
@@ -2,7 +2,7 @@
 
 ## What is Axios?
 
-Axios is a JavaScript framework that enables developers to send a variety of HTTP requests, performs automatic and contextual serialization of response data, protects
+Axios is a JavaScript library that enables developers to send a variety of HTTP requests, performs automatic and contextual serialization of response data, protects
 against [XSRF](https://en.wikipedia.org/wiki/Cross-site_request_forgery) attacks, and even allows for request cancellation amongst much more. It can be used in both
 the backend and frontend of a full stack application, making it a versatile addition to any JavaScript based application.
 
@@ -53,6 +53,14 @@ Axios.post("https://www.website.com", someData).then(
   }
 )
 ```
+When using Axios.post(), the `someData` parameter could represent any kind of data that needs to be sent in the request body, such as a JSON object.
+In addition, `res` is a JSON object that encapsulates the entire response and `err` is a similar JSON object for the error. Usually, `res` has the following attributes:
+
+* `res.status`, the HTTP response code such as 404, 200, 500, etc.
+* `res.statusText`, the text of the HTTP response code such as 'OK', 'PAGE NOT FOUND, 'INTERNAL SERVER ERROR', etc.
+* `res.data`, the actual information request contained in a JSON format. The format can vary depending on the backend/server response.
+* `res.headers`, the headers of the response that usually instruct the browser or receiver to perform a certain protocol/instruction.
+* `res.request`, is the is the request that generated this response. It is the last ClientRequest instance in node.js and an XMLHttpRequest instance in the browser.
 
 Along with `.get()` and `.post()`, Axios also supports `.put()`, `.patch()`, `.delete()`, and the entirety of the HTTP request frameworks.
 
--- a/Topics/Tech_Stacks/Axios.md
+++ b/Topics/Tech_Stacks/Axios.md
@@ -1,0 +1,78 @@
+# Axios: An all-in-one HTTP tool
+
+## What is Axios?
+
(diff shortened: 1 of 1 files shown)


Topics/Tech_Stacks/Axios.md
```diff
@@ -0,0 +1,78 @@
+# Axios: An all-in-one HTTP tool
```
JerryZhao1025, line 1: Overall, the paragraph provides a clear and concise introduction to Axios, highlighting its features and providing examples of how it can be used. However, there are a couple of areas where it could be improved:

Topics/Tech_Stacks/Axios.md
```diff
@@ -0,0 +1,78 @@
...
+
+## What is Axios?
+
+Axios is a JavaScript framework that enables developers to send a variety of HTTP requests, performs automatic and contextual serialization of response data, protects
```
JerryZhao1025, line 5: Axios is actually a library, not a framework. Frameworks provide a full structure for building applications, while libraries provide specific functionality that can be incorporated into an application. So, it would be more accurate to refer to Axios as a "JavaScript library".

Topics/Tech_Stacks/Axios.md
```diff
@@ -0,0 +1,78 @@
...
+
+```
+yarn add axios
+```
+
+Once installed, make sure to include the following at the top of any .js file that makes use of axios:
+
+``` JavaScript
+const axios = require('axios');
+```
```
JerryZhao1025, line 26: It is nice and clear.

Topics/Tech_Stacks/Axios.md
```diff
@@ -0,0 +1,78 @@
...
+    console.log(res.data);
+  }
+).catch(
+  function(err){
+    console.log(err);
+  }
+)
+```
+
+Along with `.get()` and `.post()`, Axios also supports `.put()`, `.patch()`, `.delete()`, and the entirety of the HTTP request frameworks.
```
JerryZhao1025, line 57: You could add a little bit more details or explanations below to the examples. For example, it would be helpful to explain what the res object is, where it comes from, as well as err. Those things look straightforward for us who know JS, promise and APIs. But it might be challenging for someone who learns this first day.

Topics/Tech_Stacks/Axios.md
```diff
@@ -0,0 +1,78 @@
...
+    username: 'User1',
+    password: '12345'
+  }, {
+    headers: {
+      'Content-Type': 'multipart/form-data'
+    }
+  }
+)
+```
+Axios also supports `async` and `await` so that we can wait for the necessary data and for the state of the connection to be valid.
```
JerryZhao1025, line 72: Combine with the then part (result part) to make the whole request loop clear, starting from sending the request, and ending on receiving data back.

Topics/Tech_Stacks/Axios.md
JerryZhao1025, line 57: FYI, the result you received is not just the data backend sent, it includes:
where ppl usually need to check res.status first and then use res.data to continue
{
  // `data` is the response that was provided by the server
  data: {},
 
  // `status` is the HTTP status code from the server response
  status: 200,
 
  // `statusText` is the HTTP status message from the server response
  statusText: 'OK',
 
  // `headers` the headers that the server responded with
  // All header names are lower cased
  headers: {},
 
  // `config` is the config that was provided to `axios` for the request
  config: {},
 
  // `request` is the request that generated this response
  // It is the last ClientRequest instance in node.js (in redirects)
  // and an XMLHttpRequest instance the browser
  request: {}
}

Topics/Tech_Stacks/Axios.md
BKPrograms: Thanks!

Topics/Tech_Stacks/Axios.md
BKPrograms: Good point, I've explained this along with what res can contain

Topics/Tech_Stacks/Axios.md
BKPrograms: I meant to make this a separate section from the previous part, so I'm not sure if combining it would be the best thing to do?

Topics/Tech_Stacks/Axios.md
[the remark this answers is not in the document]
BKPrograms: Added this on in latest commit

BKPrograms: This is really good information, I've added it in right below the POST request section

Topics/Tech_Stacks/Axios.md
JerryZhao1025: Well, it is really optional. I think what you have done so far should be good enough.

PR #98, opened 2023-10-10
Adding trunk based development article to website

WeissSa:
This PR adds a short description of trunk-based development as well as links/descriptions of two separate articles which discuss it. The first article is by Atlassian and does a great job describing the benefits of trunk-based development. The second is by an organization called Trunk Based Development and gives a great description of describing the Branch by Abstraction strategy.


Topics/Development_Process/Trunk_Development.md
```diff
@@ -0,0 +1,5 @@
+### Trunk-Based Development
+
+Knowing how to manage the creation of large new features in your repository is an important skill, especially when working in large organizations or on projects with complete CI systems. It can be tempting to have large feature branches with hundreds of new lines of code and it can seem daunting to implement a large feature in small stages. Trunk-based development is a system to handle large changes like this. It focuses on creating small PR and using tools such as feature flags. The following article from Atlassian does a great job discussing all the benefits of trunk-based development (i.e. impact on automated testing, having fewer active branches, and more): **[Atlassian Article](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development)**
+
+Instead of using feature flags, one can also use the "branch by abstraction" method found in [this article](https://trunkbaseddevelopment.com/branch-by-abstraction/). This is similar to referencing an interface instead of a class and much like dependency inversion, can both increase or reduce the complexity of your code.
```
williamyao27, line 5: Since feature flags are quite a useful tool and we're even using them in our own partner project, would you be able to write a short section in between your first and second paragraph describing how feature flags can be used/implemented in trunk-based development? Maybe a concrete example of how they help us break down the implementation of a "post on social media" feature into digestible batches and what type of flags may be derived in that example (doesn't have to be social media, just the first thing that came to mind).

And then the transition into the next paragraph `Instead of using feature flags,` I think will have a bit more meaning as now they can contrast the different methods of breaking down a large feature.

williamyao27, line 5: Maybe you can do a one-sentence elaboration on which situations to use it, since it can be a double edged sword in that in certain cases it may actually increase the complexity of your code.

williamyao27 (approved):
Looks great! To be perfectly honest, I had never explicitly read the details of trunk development but its general principles match a lot of what we've learned in CSC207/301 as well as the way we've been organizing tasks on our partner project right now.

Besides the comments I've left, I'm wondering if there's a way to expand on your explanation of the trunk-based development model with a diagram as well as a counterexample of what might happen if one doesn't follow this system. As it is right now, it's a bit abstract and I think relies on the reader implicitly knowing branching structures and the true danger of excessively large branches.

Something like the below diagram that I pulled from online (and would need to be cited) could work:
image

Topics/Development_Process/Trunk_Development.md
```diff
@@ -0,0 +1,5 @@
+### Trunk-Based Development
+
+Knowing how to manage the creation of large new features in your repository is an important skill, especially when working in large organizations or on projects with complete CI systems. It can be tempting to have large feature branches with hundreds of new lines of code and it can seem daunting to implement a large feature in small stages. Trunk-based development is a system to handle large changes like this. It focuses on creating small PR and using tools such as feature flags. The following article from Atlassian does a great job discussing all the benefits of trunk-based development (i.e. impact on automated testing, having fewer active branches, and more): **[Atlassian Article](https://www.atlassian.com/continuous-delivery/continuous-integration/trunk-based-development)**
```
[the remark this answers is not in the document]
WeissSa: Great idea!

Topics/Development_Process/Trunk_Development.md
WeissSa: Great idea! I will ad a more concrete example of how one might use them :+1:

WeissSa: Will do! I am going to modify this idea a little by just converting the feature-flag method to abstraction and then saying the developer can choose which is better (because they can). Since this topic is not the focus, it will be relatively low detail.

WeissSa:
@williamyao27 Let me know what you think of the example. I tend to lean towards ambiguity so I am a little worried that a new feature flag enthusiast might not understand.

Topics/Development_Process/Trunk_Development.md
```diff
@@ -0,0 +1,25 @@
...
+
+#### Feature Flags
+
+As mentioned above, feature flags are a fantastic tool to use when engaging in trunk-based development. The article above explains them abstractly, but a more concrete example might be:
+
+- Let's say you are working on a deployed website. You want to change the search algorithm but don't want to remove your old algorithm until your new one is finished. Because you are an excellent trunk-based developer, you decide to use feature flags!
+
+- You create your feature flag and set it to false. Then when coding your changes, you make sure that when the feature flag is false, all behaviour on your website remains the exact same. However, when you set your feature flag to true, your new feature becomes enabled.
+
+- Once your new search engine capability is fully ready, tested, and customer-approved, you enable your feature flag on deployment, and everyone gains access to it.
```
williamyao27, line 17: Does it make sense to add a line here about how once this is done, we are safe to remove the previous code relating to the old algorithm?

williamyao27 (approved):
>@williamyao27 Let me know what you think of the example.

New section looks good and I appreciate the search example since that ties into what I was writing. I added a small comment just to wrap up the explanation a bit.

The diagram is well-selected too. I've heard a lot of anecdotes from new student software teams expressing frustration about team members not using pull requests and sometimes even committing directly to main. I'm hoping this model can really help them learn the benefits of breaking down large features into bitesize tasks and creating PRs often!

Topics/Development_Process/Trunk_Development.md
WeissSa: Yes absolutely!

williamyao27:
Thanks for adding the clarification.

I think you should be good to merge! Doesn't look like there are any merge conflicts. Great work again @WeissSa :)

WeissSa:
@williamyao27 great reviewing!

PR #105, opened 2023-11-09
Create StripeAPI.md File

VinayakMaharaj:
Added in the learning StripeAPI in JS environment topic file


tejcapildeo:
I think your contribution is well put together and quite thorough in ensuring that all the steps for installing and setting up Stripe are included. I like that you even included information about keeping the integration secure and and testing. However I think that maybe it would be easier to follow if the subsections had more detailed descriptions summarizing what they are about. Of course some sections are straightforward and won't need much context and descriptions, but it would help to include more details in some of the sections.

tejcapildeo:
I reviewed the changes, and I feel like it's now a bit more clear what the whole thing is about. Giving a descriptive summary about Stripe as well as highlighting the pros and cons will definitely serve as something useful that makes this document more digestible.

VinayakMaharaj:
Thanks for the feedback, and suggestions. Can you also clarify if I should expand upon other sections in the readme

tejcapildeo:
Yeah I think that could definitely be a good idea. Specifically, I think you can briefly discuss the significance of a few sections. For the section on obtaining API keys, you can discuss what they are used for, and emphasize that they should be kept safe. You can also talk about the significance of handling webhook events. I think these might be helpful for providing context, especially for someone new to web development.

VinayakMaharaj:
Thanks for the feedback, you are right about needing to provide more insight there, feel free to look at the changes made

Dario1031:
Very comprehensive contribution! I think that your document would benefit from a diagram to provide a supporting example to the webhook explanation. Something like this 
image

VinayakMaharaj:
ok thanks for the feedback, implemented it and going to merge now!

PR #106, opened 2023-11-09
Adding Introductory description to Lean Methodology

ThomasShimUoft:
Created Lean.md in Software_Engineering directory and added a Lean bullet point in Software_Engineering.md


DaneshKohina:
This looks good Thomas. However I do have some suggestions to make. The main suggestion is to add a section that talks about the main differences and similarities to other popular development methodologies such as agile or waterfall? Also why don't you add a simple example workflow so that students can see the lean methodology in use. Finally also consider adding one or two more disadvantage to the lean methodology so that you have an even amount of advantages and disadvantages.

ThomasShimUoft:
Okay thank you for the feedback will add those points.

ThomasShimUoft:
I have added more disadvantages to the Lean methodology, I have also added differences between lean, agile and waterfall. There is also already a simple example workflow (although it is high overview), this framework that is mentioned in the commit is sufficient.

DaneshKohina (commented):
This is great work Tom.
Thanks for taking my changes into account. I like the addition of a Lean methodology this will definitely help future students.

PR #107, opened 2023-11-09
Adding Fuzzy Search article to wiki

williamyao27:
This PR adds a new article to the wiki discussing fuzzy search, its possible implementations, its importance to user experience, and its various limitations. Since most applications have some sort of use case for searching (e.g. searching for friends on social networks, products on e-Commerce platforms, listings on online catalogues, etc.), this article should be useful to most developers seeking to implement a non-semantic search bar in their projects.

I have placed this article under "User Experience" since besides it being a neat programming concept, it can really make-or-break the usability of an application depending on how often the user needs to search for something on it.

The article I added contains links to both tutorials on implementing fuzzy search as well as libraries that do so more optimally depending on which tech stack the reader is interested in (with a focus on JS projects).

Besides the article itself, this PR modifies the main section of the "User Experience" topic on the wiki so that it properly links to the article.


Topics/User_Experience/Fuzzy_Search.md
```diff
@@ -0,0 +1,52 @@
...
+ 
+## Implementation: Levenshtein distance
+In this section, we discuss the [Levenshtein distance](https://en.wikipedia.org/wiki/Levenshtein_distance) and how to use it in implementing fuzzy search.
+
+Levenshtein is a popular distance metric to compare the similarity between a query string and a target string. Intuitively, it counts the fewest number of times a character in the query must be deleted, inserted, or replaced to reach the target. For instance, if the query is `Brian`, its Levenshtein distance to `Bryan` is 1 as it only needs to replace `i` with `y` to reach the target. Notice that this can also be done in two operations by deleting `i` and inserting `y`, but this is suboptimal. Targets with a lower Levenshtein score are considered closer to the query.
+
+The algorithm for computing Levenshtein distance is not overly complicated and can be a fun exercise in designing recursive functions. [This GeeksforGeeks article](https://www.geeksforgeeks.org/introduction-to-levenshtein-distance/) describes several possible implementations in different programming languages should you decide to code it yourself.
```
WeissSa, line 26: Instead of phrasing it as a fun exercise, could it be worth re-phrasing it to suggest the article as a way of understanding the algorithm and then offering the option of coding one's own version with a twist (i.e. it places an emphasis on keeping the same word length instead of counting adding/subtracting a letter as equal weight as replacing one)

Topics/User_Experience/Fuzzy_Search.md
```diff
@@ -0,0 +1,52 @@
...
+
+That being said, you can also find out-of-the-box solutions online that are likely better optimized. For applications built in JavaScript, the [js-levenshtein](https://www.npmjs.com/package/js-levenshtein) library is quite efficient and easy to use:
+- Install it by running `npm install --save js-levenshtein`.
+- In your code, import the module using `const levenshtein = require('js-levenshtein');`
```
WeissSa, line 30: Since most projects use js modules, could it be worth specifying to use either require('js-levenshtein') or import {x} from 'js-levenshtein'

Topics/User_Experience/Fuzzy_Search.md
```diff
@@ -0,0 +1,52 @@
...
+The following code snippet sorts a list of names by their Levenshtein proximity to `Brian`:
+```
+const query = 'Brian';  // Define the query string
+const names = ['Bryan', 'Brian', 'Bob', 'Bryant', 'HarrisonFord'];  // Define the list of strings to search from
+names.sort((a, b) => levenshtein(query, a) - levenshtein(query, b));  // Sort the list based on Levenshtein distance
+
+// Result: ['Brian', 'Bryan', 'Bryant', 'Bob', 'HarrisonFord'] with distances [0, 1, 2, 4, 9] respectively
+```
+
+The above strategy can be used, as an example, to take in an input query string from a search bar component and sort the list of results to display before they are rendered back to the user. Additionally, instead of just sorting by distance, you can devise a threshold so that only results below a certain distance are included, to prevent extremely distant results like `HarrisonFord` from appearing at all, and those that do appear are sorted as before.
```
WeissSa, line 42:
```suggestion
The above strategy can be used, as an example, to take in an input query string from a search bar component and sort the list of results to display before they are rendered back to the user. Additionally, instead of just sorting by distance, you can devise a threshold so that only results below a certain distance are included, to prevent extremely distant results like `HarrisonFord` from appearing at all.
```
You could probably remove the last phrase. (nitpick)

WeissSa (approved):
Very detailed work! I have some small suggestions which are below this comment. I really appreciate all the excellent examples.

Topics/User_Experience/Fuzzy_Search.md
williamyao27: You're right - that's probably a better way of phrasing it. Just one thing - could you clarify if you mean the twist is coming from the article? From what I can see the implementations treat all of the operations as equal weight. Or do you mean that by coding on one's own, they can make such a customization to the algorithm whereas an out-of-the-box solution wouldn't afford them that flexibility?

Topics/User_Experience/Fuzzy_Search.md
williamyao27: Sure, will do. Actually the interesting thing is that as I was writing that I wasn't sure if it was extraneous, so this is a good confirmation that it should be removed. Thanks!

Topics/User_Experience/Fuzzy_Search.md
williamyao27: Great catch, I'll make that change as well.

Topics/User_Experience/Fuzzy_Search.md
WeissSa: I meant the latter! Thank you for clarifying

williamyao27:
@WeissSa I was wondering if you had any thoughts on the Motivation section where I describe the connection to User Experience. I wasn't exactly sure which topic the subject of search falls under since specific algorithms that are independent of a tech stack don't seem to have a place on the wiki. I'm hoping that the section I wrote can help justify why it's under the User Experience topic.

Topics/User_Experience/Fuzzy_Search.md
williamyao27: Sure, I'll write that in. Great idea!

WeissSa:
@williamyao27 I thought the motivation section did a great job connecting fuzzy search to UX. I loved how you brought in the seven universal principles of design! I think any future reader of your section would appreciate the explicate explanation and connection.

williamyao27:
@WeissSa Just seeking a final check. I don't have write permission to the repo as of now but would I be good to merge as soon as I get it? Let me know, thanks!

WeissSa:
@williamyao27 Absolutely! I think this PR will help many future generations of CSC301 students.

PR #110, opened 2023-11-10
tech-stack: added NgRx state management to Angular.md

chenikabukes:
This PR contains:
- An introduction to state management in Angular using NgRx state management
- Installation requirements for NgRx
- Key parts of NgRx state
- NgRx Benefits
- NgRx Difficulties

### DIFF
--- a/Topics/Tech_Stacks/Angular.md
+++ b/Topics/Tech_Stacks/Angular.md
@@ -8,7 +8,10 @@
 Angular is a framework which allows people to create responsive single page web apps. It is built on TypeScript and can be used to build scalable web apps, has a collection of well-integrated libraries which cover many features and several tools to help build, test and maintain your code. With Angular you can create single-page-application where you have a single html file where changes are rendered in the browser. Familiarity with HTML/CSS as well as JavaScript/TypeScript will be needed while building your web app.
 
 ## NgRx State Management
-Angular applications often deal with complex state management as they grow in size and complexity. NgRx is a well-organized state management library for Angular applications that leverages reactive programming concepts. It is inspired by Redux, a popular state management library for React applications, but is more structured (in file structure and component interaction). You can get started with NgRx by using this npm command: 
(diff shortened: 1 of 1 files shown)


anudevgill:
This looks great and is very informative! However, I would recommend adding examples of each of the key parts of the state to make it easier for someone with less front-end experience to understand.

chenikabukes:
> This looks great and is very informative! However, I would recommend adding examples of each of the key parts of the state to make it easier for someone with less front-end experience to understand who may not have actually worked with NgRx state management before. These examples may make the concepts more relatable and easier to digest/understand.

Thank you for pointing that out Anudev! I will add some simple examples to each part of state to make the comprehensive state management system which is NgRx more clear.

RJAK11:
This is really well-organized and easy to follow! This sounds like a very comprehensive state management system, can you add a brief explanation about if there are possible simplifications for when the database of your application gets too large to individually handle each update?

chenikabukes:
> This is really well-organized and easy to follow! It sounds like a very comprehensive state management system, can you add a brief explanation about if there are possible simplifications for when the database of your application gets too large to individually handle each update?

Thanks for the feedback Rana! That's a great point, I will add an explanation about how we can use memoization in state to resolve this problem.

chenikabukes:
@RJAK11 @anudevgill Hi reviewers! I have addressed your comments above and would like to confirm if there is anything else I should address, otherwise are we ready to merge?

RJAK11:
Yup, that looks perfect to me. Great Job Chenika!

RJAK11 (approved):

anudevgill:
> @RJAK11 @anudevgill Hi reviewers! I have addressed your comments above and would like to confirm if there is anything else I should address, otherwise are we ready to merge?

It looks great! The code snippets are particularly helpful in understanding the content!

anudevgill (approved):
Looks great! You can go ahead and merge!

PR #114, opened 2023-11-14
What is Url sanitization

WillJarvis-Cross:
This document gives an oversight of what URL sanitization is, why to use it, how to implement it, and it gives an example of a possible implementation.

### DIFF
--- a/Topics/Development_Process.md
+++ b/Topics/Development_Process.md
@@ -16,6 +16,9 @@
 ## React Testing Library
 ### [React Testing Library](./Development_Process/React_Testing_Library.md)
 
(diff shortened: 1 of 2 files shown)


parth2324 (changes requested):
There are are few spellings mistakes, like "learnging" in introduction. I recommend proofreading once the entire URL_Sanitization document. I like your sanitization examples, took me a minute to realize "&# 104" string ! Could you add some sample test cases for the custom sanitizing function you made. Otherwise looks good!

parth2324 (approved):
Looks perfect ! Great work !!

PR #118, opened 2023-11-16
Tech Stack: Added React useEffect with Axios

hanshhh:
This PR creates a file called React_useEffect_Axios.md for Tech Stack section.

1. It provides a beginner-friendly, step-by-step tutorial for React useEffect Hook.
2. It clarifies the concept of dependency array, which was the most confusing part when I first learned the Hook.
3. It offers code examples of using useEffect and Axios to fetch and load data in the frontend.
4. In the Additional Resources Section, I include a blog with detailed explanation of the cleanup function which learner might want to refer to if they want to learn more about the solution to memory leak. Also, there is a great in-depth tutorial from freeCodeCamp that I found useful when I was learning Axios and useEffect.

### DIFF
--- a/Topics/Tech_Stacks.md
+++ b/Topics/Tech_Stacks.md
@@ -6,6 +6,7 @@
 ### [Building Apple Native Software Using Swift and SwiftUI](./Tech_Stacks/swift.md)
 ### [How to access and use salesforce API](./Tech_Stacks/salesforce_api.md)
 ### [React Resources](./Tech_Stacks/React.md)
+### [React useEffect Hook with Axios](./Tech_Stacks/React_useEffect_Axios.md)
 ### [Angular Resources](./Tech_Stacks/Angular.md)
 ### [Learning PostgreSQL and psycopg2](./Tech_Stacks/PostgreSQL_psycopg2.md)
 ### [CSS](./Tech_Stacks/CSS.md)
--- a/Topics/Tech_Stacks/React_useEffect_Axios.md
+++ b/Topics/Tech_Stacks/React_useEffect_Axios.md
@@ -119,11 +119,14 @@ The code above is a simplified example to help you understand the proper way of 
 
 ## Additional Resources
 The offical documentation for useEffect Hook:
+
 https://react.dev/reference/react/useEffect
 
 A blog post that goes into depth about the cleanup function:
+
 https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/
 
 FreeCodeCamp Tutorial about Axios in React:
+
 https://www.freecodecamp.org/news/how-to-use-axios-with-react/
  
--- a/Topics/Tech_Stacks/React_useEffect_Axios.md
+++ b/Topics/Tech_Stacks/React_useEffect_Axios.md
@@ -1,12 +1,18 @@
 # Learning UseEffect Hook with Axios
 
 ## Table of Contents
-### [Introduction](Introduction-1)
-### [What is useEffect Hook?](What-is-UseEffect-Hook?-1)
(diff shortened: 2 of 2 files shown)


KareemSalem7 (changes requested):
This is a good breakdown for React useEffect. However, it would be better if for dependency array there was an example to clarify that concept clearly with code rather just plain text.

hanshhh:
> This is a good breakdown for React useEffect. However, it would be better if for dependency array there was an example to clarify that concept clearly with code rather just plain text.

Thanks for pointing that out. When I was writing it, I tend to make it as brief and simplified as possible so that everyone would understand. But it's true that adding an example for the most confusing dependency array concept would make things much clearer.

Kai15116 (changes requested):
Good job! Overall, it looks great, but I found a few improvements that you might be able to make. I actually like how you added an example for loading since that's a common function to add for the web. 
- The table of contents links seem to not be working. I think you are missing # in the bracket for each link in the table of contents. 
- I really like how you added comments in the first three example codes. So, I think it would be better if you added comments for the last code example to make it easier for people to understand your example. 
- One final suggestion is to add a prerequisite at the top. I think it will help people to know what is the background knowledge required to understand this material. 

Except for these, I didn't find any issues with your content. Great work!

hanshhh:
> Good job! Overall, it looks great, but I found a few improvements that you might be able to make. I actually like how you added an example for loading since that's a common function to add for the web.......

Oops, my bad, I'll fix the links. Thanks for the catch. By the way, Do u think I should have more detailed example with empty dependency array?

Kai15116:
> > Good job! Overall, it looks great, but I found a few improvements that you might be able to make. I actually like how you added an example for loading since that's a common function to add for the web.......
> 
> Oops, my bad, I'll fix the links. Thanks for the catch. By the way, Do u think I should have more detailed example with empty dependency array?

I think your first example suffices for most cases, especially because empty dependency arrays are not often used. However, you might be able to add a small failing example for the "infinite loop of requests'" that you mentioned, so readers don't make the same mistakes.

hanshhh:
@Kai15116  Got it!

Kai15116 (changes requested):
The changes looks great.
The table of contents links are now working and comments on the axios code example looks good. You also illustrated failing example at the bottom which is also good. All my requested changes seems to be done. 

I might be nitpicking a bit, but I think "useState Hook" can be good topic to add for the prerequisite since you are using them throughout  your content.

KareemSalem7 (changes requested):
This is an improved version I like the inclusion of code examples that really illustrate the use of use effects.

Here are some improvements
-  There is somewhat of a redundancy of arrow function try to reduce this and make it comprehensive.
-  If possible, include a source for explaining mount and unmount so that users can go to that source and investigate the documentation themselves. This can help with long term comprehension and trust of your information.

hanshhh:
In my latest commits, 
@Kai15116 I updated the prerequisites, added comments and common mistakes for the Axios Demo.
@KareemSalem7 I removed code example redundancy and included sources for life cycle methods.

Thanks for all the feedbacks and suggestions.

Kai15116 (approved):
All the changes looks good! Prerequisite are also updated. 
Good job!

KareemSalem7 (approved):
This is great you walk the user step by step on the process of using useEffect with axios, the user can understand the impact of it's use, examples to guide the user, common mistakes that help people understand what to avoid, and sources they can do further reading on to expand their knowledge

KareemSalem7:
@hanshhh hi hans you are my most trusted confidant. I need your advice as a expert programmer. Should I include a conclusion to recap the report?

PR #120, opened 2023-11-17
System Design: Twitter Case Study

ericjest:


Topics/System_Design/Twitter_System_Design.md
```diff
@@ -0,0 +1,63 @@
...
+
+Twitter's initial naive approach was to join the Tweet, User, and Follows tables in SQL upon a GET request from a client like so:
+```
+SELECT * FROM tweets
+JOIN users ON tweets.author_id = users.id
+JOIN follows ON follows_id = users.id
+WHERE follower_id = current_user.id;
+```
+
+You can quickly see how this approach became quite slow at over 300K QPS.
```
k-m-crawford, line 44: Elaborate on this maybe?

Topics/System_Design/Twitter_System_Design.md
```diff
@@ -0,0 +1,63 @@
...
+
+![Twitter's Caching Approach](twitter.png)
+*A system diagram of Twitter's caching approach for home timelines.*
+
+## A Third Twist: the Hybrid Approach
+The caching approach worked for accounts with a reasonable number of followers. Assuming an average follower count of 100 and 4.6k Tweets per second (based on November 2012 data), this translated to 460k writes per second to the Redis cluster.
+
+However, for celebrity accounts with huge numbers of followers, this was not performant. A Tweet from Lebron James, who has 50 million followers as of 2023, would result in over 50 million writes at once. In this scenario, the caching approach was not feasible given Twitter's goal of delivering Tweets within 5 seconds. As a result, Twitter implemented a hybrid approach, where most new Tweets were cached (using approach 2) while Tweets from accounts with large followings were queried at runtime (like approach 1) and merged chronologically with the Tweet cache.
```
k-m-crawford, line 57: What is the benefit to the hybrid approach for large accounts? Wouldn't the same issues exist for approach 1 with large celebrity accounts?

Topics/System_Design/Twitter_System_Design.md
```diff
@@ -0,0 +1,63 @@
...
+SELECT * FROM tweets
+JOIN users ON tweets.author_id = users.id
+JOIN follows ON follows_id = users.id
+WHERE follower_id = current_user.id;
+```
+
+You can quickly see how this approach became quite slow at over 300K QPS.
+
+## Approach 2: Caching
+As a result, Twitter implemented a fan-out caching system. They stored a [Redis list](https://redis.io/docs/data-types/lists/) for each user containing the Tweets to display on each user's home timeline. Let's illustrate this with an example. Say that Darth Palpatine is followed by Anakin Skywalker, Obi-Wan, and Padme Amidala. When Darth Palpatine posts a Tweet, Twitter's fan-out service pushes the new Tweet to the Redis cache for each user that follows Darth Palpatine (Anakin, Obi-Wan, and Padme). This way, when Anakin (for example) loads his timeline page, his Tweets are cached in Redis and can be retrieved in constant runtime (in this simplified example).
```
k-m-crawford, line 47: did they use Redis lists for a reason vs. other data structures or libraries? This isn't necessarily an issue with the write up here just curious / maybe worth comparing other possible approaches to elucidate why this is the most logical solution. perhaps could briefly examine other possible solutions?

Topics/System_Design/Twitter_System_Design.md
k-m-crawford: Also is this hybrid approach employed at a certain amount of followers? Curious what the exact criteria may be (i.e., what if someone has 1 million followers vs 50 million?)

Topics/System_Design/Twitter_System_Design.md
k-m-crawford: maybe also a quick blurb on what Redis is for those unfamiliar

Topics/System_Design/Twitter_System_Design.md
ericjest: Do you think it's unclear? I felt like it was evident that JOINing the entire userbase with all Tweets and all follow relationship for every query would not be a good idea. But maybe that isn't clear, should I add that explanation?

Topics/System_Design/Twitter_System_Design.md
ericjest: To understand why the hybrid approach is better, you can broadly think about the advantages and disadvantages of approach 1 and 2.

Approach 1's complexity increases as the size of our Users, Tweets, and Followers tables gets really large. However the catch is that Twitter users don't really follow millions of users, so we only have to consider Users and Tweets.

Approach 2's complexity increases based on how many accounts follow you.

We've established that Approach 2 isn't the best with celebrities. Now let's consider the benefits of the hybrid approach over Approach 1.

We can create tables CelebrityTweets and CelebrityUsers. When a user queries their homepage, we get from the Redis cache in O(1) runtime. Then we JOIN CelebrityTweets and CelebrityUsers to get any Tweets from celebrity accounts they might follow. Note that the size of CelebrityTweets and CelebrityUsers is quite limited relatively (to our original approach).

Of course, given the popularity of Twitter, this hybrid approach might still have scalability issues (this is why scalability is a complex issue). But the point of the ultra-simple example in this article is that the hybrid approach is much better than _just_ doing Approach 1.

Do you think this explanation makes sense? I can add it to the article if you feel like it would be a good addition.

ericjest: And not sure about exact follower cutoffs. I think that's out of the scope of this article and probably a private implementation detail to Twitter.

Topics/System_Design/Twitter_System_Design.md
ericjest: I think any in-depth discussion on Redis is out of the scope of the article, but I can add a short explanation on what Redis is, as well as a link to an article with some reasons that Twitter chose Redis.

Topics/System_Design/Twitter_System_Design.md
k-m-crawford: I guess it depends on who's reading. "You can quickly see how [JOINing the entire userbase with all Tweets ...] became quite slow at over 300K QPS"? Probably not a major issue since it is clear why if you have basic database knowledge

Topics/System_Design/Twitter_System_Design.md
k-m-crawford: I think the explanation is well-written; worth adding I think.

k-m-crawford (approved):

PR #121, opened 2023-11-17
Adding HTML guide to A2 class wiki contribution.

colewiltse:
Here, I am adding the first version of my A2 contribution. I wrote a short guide to the basics of HTML to help make it easier for others to get started writing their html documents.

### DIFF
--- a/Topics/Tech_Stacks/HTML_Guide
+++ b/Topics/Tech_Stacks/HTML_Guide
@@ -1,0 +1,75 @@
+# HTML Guide
+
+## Table of contents
+### [Introduction](#introduction-1)
+### [Syntax](#syntax-1)
+### [Elements](#elements-1)
+### [Writing HTML Files](#writing-html-files-1)
+
+## Introduction
+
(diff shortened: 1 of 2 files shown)


PatrickFidler (changes requested):
This is a great guide on the basics of HTML! I think the link to other html tags will be especially helpful for learners who need to frequently reference the various tags.
When I was reading the section about writing/testings HTML files, I felt that if I didn't know what vscode was, I would be a bit confused about its purpose and what the extension is actually "extending". Could you add a link to vscode's home or download page? Or perhaps just mention it's role in the testing.
I also noticed that the file itself is missing the `.md` extension, so github isn't showing the markdown.

colewiltse (commented):
Thank you for the feedback! I didn't even notice I forgot to save my file as a .md file. I have just fixed the file type and have also provided further information on vs code for those unfamiliar with it.

PatrickFidler (approved):
Looks great! The page provides a nice summary on the basics of HTML, and directs readers to good resources to learn more.
The section on testing your HTML is especially useful; that vs code extension makes making minor editing to the webpage very easy, and doesn't overcomplicate things too much for beginners.

PR #144, opened 2023-11-20
Basics of Bootstrap

charlietaoxl:
Goal: Giving an introduction of Bootstrap to students that are unfamiliar with CSS Frameworks.

- Explain what Bootstrap is and why it is used
- Show how some Bootstrap components work, as well as the workflow
- Introduce Bootstrap Grid layout
- Show code examples that students can work off of

### DIFF
--- a/Topics/Tech_Stacks/Bootstrap.md
+++ b/Topics/Tech_Stacks/Bootstrap.md
@@ -2,7 +2,10 @@
 
 ## Table of contents
 ### [Introduction](#what-is-bootstrap?-1)
-### [References](#references-and-resources-1)
+### [What is Bootstrap](#what-is-bootstrap-1)
+### [Where to begin](#where-to-begin-1)
+### [Bootstrap Basics](#bootstrap-basics-1)
+### [References](#references-1)
 
 ## Preface
 Here you will learn the basics of how to incorporate Bootstrap into your website-building workflow. This should not serve as an in-depth guide, but rather one that should help you get your foot in the door to begin understanding and using Bootstrap components. Before you start, it is recommended that you have working knowledge of both HTML and CSS. This could be as little as a simple styled website. Although not required, it will be helpful in understanding why Bootstrap is such a widely-used and useful framework when working with CSS. At the end of this guide, you should be able to apply some simple Bootstrap components or formatting into your website.
@@ -20,7 +23,7 @@ For starters, you will need to add Bootstrap to your website's HTML files. Follo
 3. For Bootstrap's JavaScript, insert `<script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/js/bootstrap.bundle.min.js" integrity="sha384-C6RzsynM9kWDrMNeT87bh95OGNyZPhcTNXj1NW7RuBCsyN/o0jlpcV8Qyq46cDfL" crossorigin="anonymous"></script>` inside your `<body> </body>` element, at the bottom. This allows the use of certain Bootstrap elements that rely on JavaScript. 
 
 For the sake of this tutorial, insert all the above and it should look something like: 
-```
+```html
 <!doctype html>
 <html lang="en">
   <head>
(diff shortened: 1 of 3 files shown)


Topics/Tech_Stacks/Bootstrap.md
```diff
@@ -0,0 +1,163 @@
+# Learning Bootstrap, a CSS Framework
+
+## Table of contents
```
Yasutakahi, line 3: It is good that it has table of contents so it is easier for the readers

Topics/Tech_Stacks/Bootstrap.md
```diff
@@ -0,0 +1,163 @@
...
+### [Introduction](#what-is-bootstrap?-1)
+### [What is Bootstrap](#what-is-bootstrap-1)
+### [Where to begin](#where-to-begin-1)
+### [Bootstrap Basics](#bootstrap-basics-1)
+### [References](#references-1)
+
+## Preface
+Here you will learn the basics of how to incorporate Bootstrap into your website-building workflow. This should not serve as an in-depth guide, but rather one that should help you get your foot in the door to begin understanding and using Bootstrap components. Before you start, it is recommended that you have working knowledge of both HTML and CSS. This could be as little as a simple styled website. Although not required, it will be helpful in understanding why Bootstrap is such a widely-used and useful framework when working with CSS. At the end of this guide, you should be able to apply some simple Bootstrap components or formatting into your website.
```
Yasutakahi, line 11: Here, one of the thing I suggest and think good idea that you add some tutorials about CSS and HTML. This does not have to be your tutorial, it can be outer source that teaches you some basic knowledge about CSS and HTML. Moreover, you can briefly explain why some basic knowledge about CSS and HTML are needed or recommended

Yasutakahi (changes requested):
Overall I think it is very good as it has some actual codes and some examples to deepen the understanding about this topic. Also constantly refer to the documentation, so that the reader do not simply copy and paste your work, they will also refer to the documentation which makes the knowledge deeper. To make this better, I would suggest you to make some changes to the format. More specifically, I would suggest you to create some list of URLs for each documentation in addition to in text URL of documentation which is what you already have. This allows a user who finished reading your instruction, but want to refer to the documentation for their work or project can easily refer to those. I also asked some changes in the comments so please refer to those as well as this comment when you modify it!! but overall it looks very good!! keep working

Topics/Tech_Stacks/Bootstrap.md
```diff
@@ -0,0 +1,163 @@
...
+
+## What is Bootstrap?
+Bootstrap is a CSS Framework that simplifies and standardizes much of the styling process. It does this by including much of the common styling choices as an optional pre-built class that can be attached to HTML elements, most commonly divs or spans. In a typical workflow, you would need to create the CSS for these class yourself in a .css file and import those styles. Over a large project, you might begin to realize that the styles can get messy and/or redundnant. Using Bootstrap's built-in styles would instead allow you to rely on their basic foundational stylistic building blocks, requiring less custom CSS from you.
```
Yasutakahi, line 14: What is the difference between css and html? what are the benefits of using bosstrap instead of other framework? why do we use those?

Topics/Tech_Stacks/Bootstrap.md
charlietaoxl: That is a good point. Since there is already a section in the wiki for CSS, I will make sure to link that section. On that note, I think I will also link the Bootstrap section in the CSS as a next step.

Topics/Tech_Stacks/Bootstrap.md
charlietaoxl: I think my guide is more for people who at least know what CSS and HTML are or can do, so I believe it might be a little unnecessary to include something like the difference between CSS and HTML I added in the preface that it is more intended for someone who can make a simple web page, and I believe this guide will only prove useful to someone who is at that level.

However, I do agree that I should put some information about other frameworks in here.

Yasutakahi: That makes sense. Thank you for the clarification!!!

Yasutakahi (approved):
All the changes look good! improved a lot and easier than before

PR #167, opened 2023-11-22
Introduction to Flask and Flask-SQLAlchemy

tejcapildeo:
This is the introduction to Flask, specifically including the usage of Flask-SQLAlchemy as the database solution. This is the first draft, as it will be reviewed by other group members and changes will be made to improve it. The aim is to explain what each technology is and its benefits, then explain the installation, and finally give some insight on how to use these technologies.

### DIFF
--- a/Topics/Tech_Stacks/Flask.md
+++ b/Topics/Tech_Stacks/Flask.md
@@ -1,0 +1,63 @@
+# Intro to Flask with Flask-SQLAlchemy
+
+
+## Introduction
+### What is Flask and why is it useful?
+Flask is a lightweight and flexible web framework for Python. Developed by Armin Ronacher, it is designed to be simple and easy to use, providing the essentials for building web applications without imposing rigid structures. Flask is often referred to as a "micro" framework because it focuses on keeping the core simple and extensible, allowing developers to choose the tools and libraries they need. 
+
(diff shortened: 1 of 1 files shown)


VinayakMaharaj:
Good job with the overview you gave on Flask-SQLAlchemy as the database solution, seems very thorough! I suggest however that you maybe include some videos in addition to the documentation linked, maybe can be an overview video on this tool or maybe even a video similar to this can work https://youtu.be/uZnp21fu8TQ?si=c46Sk4QzZCIMn9Lf

tejcapildeo:
That's a great idea. This video seemed really useful for getting into Flask-SQLAlchemy, so I went ahead and added it to the Additional Resources section. I also added a link to more documentation in that section as well.

VinayakMaharaj:
Overall it looks really nice, just to improve visibility can you use markdown formatting to bold certain key points, can you also use ordering for your steps to setup and make the links look nicer by clicking a word instead of the link directly

tejcapildeo:
I made the changes, and I think this helps improve the readability of the document. I was wondering though, do you think I should add some more about what to do if the user is getting trouble with Flask?

VinayakMaharaj:
Yes, I think that would be really helpful. Maybe you can include some helpful documentation for common errors and how to troubleshoot them.

Dario1031:
Your guide looks great! I was thinking that you could include the terminal commands to install flask and flask-sqlalchemy inside of the report itself (even though you've already linked the guides) to make it easy to copy and paste.

tejcapildeo:
That's a good idea. I just made the changes, and I agree that this will make it easy for the reader to find and reference the installation commands.

PR #175, opened 2023-11-23
Introduction to Static Typing in Python

larryworm1127:
This pull request adds information about static typing in Python. It discusses type system-related topics including:
- dynamic typing and static typing
- typing hinting in Python and how it can be utilized for static type analysis using the Mypy library
- the pros and cons of using type hints and static type checking in Python


wang-henry:
You should probably add some links to the Python official documentation for type hints so that people who want to use a larger variety of types know where to find them.

larryworm1127:
Thanks for the feedback. I have added the links to the documentations for both Python type hints and Mypy

eric-pham:
I think you should try to be more consistent with the code example to minimize confusion. For example, the first function snippet uses `pass` while the second one uses `...`.

larryworm1127:
Good suggestion! Let me change it.

wang-henry (approved):
LGTM

PR #178, opened 2023-11-24
Added Protobuf and gRPC tutorial overview in Topics/Tech_Stacks

wang-henry:
Added section that gives readers a mini tutorial on how to quickly setup a Protobuf file and use it with gRPC.

### DIFF
--- a/Topics/Tech_Stacks/Protobufs_and_gRPC.md
+++ b/Topics/Tech_Stacks/Protobufs_and_gRPC.md
@@ -1,0 +1,46 @@
+# Protobufs and gRPC
+
+## Introduction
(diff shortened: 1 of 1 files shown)


larryworm1127:
I think it would be good to include a server response section to show people what that looks like.

wang-henry:
> I think it would be good to include a server response section to show people what that looks like.

I think we should try to keep it simple and only include the client side to not confuse the readers, especially because many codebases will already have the server portion completed and developers will simply need to add an additional client.

eric-pham:
You should still include it since it would give readers a better idea on how to use gRPC overall, even if they are not making the server side.

larryworm1127:
Also, I feel like the document ends too abruptly. You should add a conclusion section to summarize what gRPC can do and how it helps developers.

wang-henry:
Thanks for the feedback. Added server-side code and a conclusion.

larryworm1127:
Also, remember to add your article on README.md and tech_stack.md files.

larryworm1127 (approved):
LGTM once the conflicts are resolved

PR #179, opened 2023-11-24
Create clean_architecture.md

Dario1031:
Added a quick crash course guide on Clean Architecture, including its main principles, how to implement it in your projects, and its advantages/disadvantages. Also included links to helpful resources.


VinayakMaharaj:
it looks really comprehensive so far, maybe to bring ur point across more easily u can use an image like what they have here
https://blog.cleancoder.com/uncle-bob/2012/08/13/the-clean-architecture.html

tejcapildeo:
I like the work you've done, and it seems to cover everything efficiently and effectively. I like that you included an example with the translation API to illustrate Clean Architecture. I think something that might be helpful to add is a brief comparison of Clean Architecture with some other common Architectures, such as Monolithic or Microservices.

Dario1031:
You're right about it being helpful to cover some common architectures as a comparison. I added some points about the ones you mentioned and made a commit.

Dario1031:
I see your point about making it bulleted to enhance clarity, added that in my most recent commit!

PR #180, opened 2023-11-24
Update PostgreSQL_psycopg2.md

PatrickFidler:
I am commiting the first version of my A2 contribution. I reworked the PostgreSQL and psycopg2 section to make it more clear on how psycopg2 worked, including how to install/run it and how to use it effectively.

### DIFF
--- a/Topics/Tech_Stacks/PostgreSQL_psycopg2.md
+++ b/Topics/Tech_Stacks/PostgreSQL_psycopg2.md
@@ -3,8 +3,8 @@
 ## Table of contents
 ### [Prerequisites](#prerequisites-1)
 ### [Introduction](#introduction-1)
-### [PostgreSQL installation](#postgresql-installation-1)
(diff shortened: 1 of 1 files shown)


colewiltse (changes requested):
The changes you have made make it much clearer to understand how to set up and use psycopg2. Specifically, the in-depth steps you updated clearly state what you are doing and why, allowing the user to understand without just following blind instructions.

I do however have some small requests. Near the end of the document you briefly mention using `fetchone()` and `fetchmany(x)`. It would be helpful to provide a real-world scenario to show why you would want to use these functions. Also, for the links you provided in the additional resources section, a short description of what these links are may be helpful to give the reader a taste of what to expect.

PatrickFidler:
Great suggestions.
- Added practical example to `PostgreSQL_psycopg2.md` showing why functions like `fetchone()` are useful
- Added descriptors to both links in the Additional Resources section to add clarity about their inclusion

PR #195, opened 2023-11-26
PO vs PM Reverted Branch

tupi2001:


ayanaar:
Your PR provides a clear and detailed comparison between Product Managers and Product Owners, which is beneficial for understanding these distinct but often confused roles. However, I have a couple of suggestions to further enhance the content:
- Including a few real-world case studies or examples where the roles of Product Managers and Product Owners made a significant difference in a project could provide practical insights. This would help readers understand how these roles function in actual business scenarios.
- Ensure consistency in terminology throughout the document. For example, if 'Product Owner' is abbreviated as 'PO' in one section, this abbreviation should be consistently used throughout the document.

tupi2001:
This is very useful feedback. Thank you for pointing out that I need to be consistent with my abbreviations. I changed it in the most recent commit. Regarding me providing real examples of PO vs PM, that is relatively difficult to do that because it is a recent field and there is often a huge overlap between the two roles. It is often the case when one company only has one of those role. I tried to explain this in a passage at the end so hopefully I made my point.

ayanaar:
How will this page be accessed by the readers on the website? Where will you link it? As a suggestion, you could provide a link to the section in `Product_Management.md` file.

ayanaar (approved):

PR #196, opened 2023-11-26
Firebase Realtime Database Contribution

mistryrohan:
My initial PR is ready to be reviewed. Please let me know anything I should add or remove for clarity.


whyismynamerudy (changes requested):
This looks good! I just had a quick suggestion: if you could also include the advantages and disadvantages of Firebase that would be good in providing everyone else with more insights using which they can base their decision to use this technology or not.

mistryrohan:
I think that is a great idea considering that there are many NoSQL options so I will add those changes in right now.

whyismynamerudy (approved):
The advantages and disadvantages look good! Thanks for adding in the requested information.

PR #198, opened 2023-11-26
Created topic + first draft

SpecialThing44:
I think this is reasonably comprehensive to provide practical information and motivate and interested reader to research further.

### DIFF
--- a/Topics/Software_Engineering/Static_Analysis_and_Verfication.md
+++ b/Topics/Software_Engineering/Static_Analysis_and_Verfication.md
@@ -1,0 +1,87 @@
+# Introduction to Static Analysis and Formal Verification
+
+## Table of Contents:
+- ### [Example Applications](#Example-Applications)
+- ### [Static Analysis](#Static-Analysis)
+- ### [Formal Verification](#Formal-Verification)
+- ### [Additional Resources](#Additional-Resources)
+___
+
+
+
+## Example Applications
+
+People write software for a variety of purposes, ranging from simple e-commerce websites to helping navigate space rockets.
+In the former example, the risks are small. A bug in the code might mean a product might not be rendered properly, or trying to access a page returns an error.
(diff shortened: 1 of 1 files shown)


jeanine5 (changes requested):
This is really well written and organized. One minor suggestion is to separate the first bit of info in the 'Example Applications' section into an Introduction section. In this section, you could briefly mention/introduce some methods you talk about later.

rowanArora (changes requested):
You have done a very good job with this breakdown! However, I believe the formatting of your writing could be better. Specifically, your "Example Applications", "Static Analysis", and "Formal Verification" sections are all written in a more typical essay way, but your other sections are all written in bullet-points, so I think if you were to just consistently stick with one format for your writing, it would look cleaner. 
Outside of the formatting, I do wish you could expand a bit more on the "Formal Verification" section as it feels quite bare at the moment compared to the rest of your sections. 
You did a great job!

SpecialThing44:
> You have done a very good job with this breakdown! However, I believe the formatting of your writing could be better. Specifically, your "Example Applications", "Static Analysis", and "Formal Verification" sections are all written in a more typical essay way, but your other sections are all written in bullet-points, so I think if you were to just consistently stick with one format for your writing, it would look cleaner. Outside of the formatting, I do wish you could expand a bit more on the "Formal Verification" section as it feels quite bare at the moment compared to the rest of your sections. You did a great job!

I agree that I could expand the intro to Formal Verification slightly, but I'd argue that the current formatting makes sense.

jeanine5 (approved):
I think this looks much better. More organized and the information flows better.

rowanArora (approved):
After rereading your work, it has been done very well! I apologize for my confusion regarding the formatting on the last review I made, I was viewing the raw code rather than the output file. I think the work you've done is structured in a more natural way and I believe your content is more comprehensive than before. Overall, you've done a great job!

PR #199, opened 2023-11-26
Add instruction on connecting to external database using Apache Superset

PenguinKuo007:
Add instruction on connecting to external database using Apache Superset


justinzzy-code:
Overall looks good, I think you can add more details. I will make a review.

justinzzy-code (changes requested):
I think you can add more details like how to set up preferred database options and images:
https://superset.apache.org/docs/databases/db-connection-ui/

Also you can consider adding details about the Extra Database Settings.

PenguinKuo007:
Added more detail on database and image

justinzzy-code (approved):

justinzzy-code:
Great commit! Added very detailed instructions outlining the process of incorporating new database engines into the existing endpoints, along with guidelines for configuring images.

DavidM0808 (commented):
I think all the set-up steps and the instruction to add new database engines are clear and detailed. However, are there any potential issues that the users can run through while setting up an database and connecting it to Superset? What if connection fails?

PenguinKuo007:
I have added instruction on potential connection failure.

Topics/Tech_Stacks/Apache_Superset.md
```diff
@@ -193,6 +193,65 @@ In addition, the DB engine spec must implement these class methods:
 
 For databases like MySQL and Postgres that use the standard format of `engine+driver://user:password@host:port/dbname`, you need to do is add the `BasicParametersMixin` to the DB engine spec, and then define the parameters 2-4 (`parameters_schema` is already present in the mixin).
 
+### Potential connection failure:
```
DavidM0808, line 196: Yep the details here is good because it demonstates what the issue stems from (Mainly Apache Superset based on what you wrote). It's also safe to assume that the database server should always be running given that you have registered for the service. Approved for merge.

DavidM0808 (approved):

PR #211, opened 2023-11-27
Design Patterns

JY3150:
I have written down a file about design patterns. I briefly talked about what they are, why we need them, and give a simple example of Factory Pattern.

### DIFF
--- a/Topics/Development_Process/Design_Patterns.md
+++ b/Topics/Development_Process/Design_Patterns.md
@@ -1,0 +1,74 @@
+# Design Patterns for Beginners
+
+## Introduction
+Design patterns are solutions to common problems encountered in software design. They provide a 'systematic' yet creative way to design and solve problems. When you're building software, you often face similar challenges, and design patterns offer a way to address these challenges in an efficient manner.
+
+### Why Design Patterns?
+
+1. **Reusability:** Design patterns promote code reusability. Once you understand and implement a design pattern, you can apply it to different projects and scenarios. Note that it is not a piece of code for you to copy and paste, but a certain way of thinking and solving problems. 
+   - But it is not the same as _algorithm_, whose input and output satisfy a certain specifications. Design Patterns are more creative and applicable in software development, in the sense that it does not require a strict specification on its input. 
+
+2. **Maintainability:** Using design patterns results in cleaner and more maintainable code. They encapsulate specific behaviors, making it easier to modify or extend functionalities without affecting the entire codebase.
+
+3. **Scalability:** Design patterns contribute to scalable and flexible software architecture. They help in building systems that can adapt to changing requirements and scale gracefully.
+
+
+## Types of Design Patterns
+
+### 1. Creational Patterns
+Creational patterns focus on object creation mechanisms, dealing with the process of object instantiation.
+
+### 2. Structural Patterns
+Structural patterns are concerned with object composition, forming relationships between objects to create larger structures.
+
+### 3. Behavioral Patterns
+Behavioral patterns define ways for objects to communicate, encapsulating patterns of communication between objects.
+
+## Simple yet Powerful Example
+We briefly introduce a **Creational Pattern** to demonstrate how powerful and flexible design patterns are. 
+
+Suppose we are developing a software where transportation tools are involved. At first, we only support cars and busses. But as our product grows, we have the need to support more types of transportation tools, like airplanes and boats, etc. We can certainly create more classes, like we did for cars and busses. But sometimes all we care about is that they are methods of transportation, i.e., they can take people from one place to another place. Let's call that behavior `move()`. 
+
+Thus, we may define a shared interface between among these classes
+```python
+class Vehicle:
+    def move(self):
+        ...
+```
+
+We can then define 
+```python
+class Airplane(Vehicle):
+    def move(self):
+        ...
+```
+and 
+```python
+class Car(Vehicle):
+    def move(self):
+        ...
+```
+
+
+This way, we don't need to worry about the exact details of 'how a car/plane/bicycle' moves, but knowing that they support `move()` is good enough. 
+
+
+## Helpful resources
+### Online Courses:
+- **Coursera - "Design Patterns" by the University of Alberta:**
+   - This online course covers design patterns using Java. It includes video lectures, quizzes, and programming assignments to reinforce the concepts.
+
+### Websites and Tutorials:
+
+-  **[Refactoring Guru - Design Patterns](https://refactoring.guru/design-patterns):**
+   - This website provides a comprehensive guide to design patterns with examples in multiple programming languages. **EXTREMELY RECOMMENDED!**
+
+
+- **[DZone - Design Patterns Refcard](https://dzone.com/refcardz/design-patterns):**
+   - DZone offers a concise reference card on design patterns, providing a quick overview of essential patterns.
+
+### Online Communities:
+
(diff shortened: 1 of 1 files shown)


Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
+# Design Patterns for Beginners
+
+## Introduction
+Design patterns are solutions to common problems encountered in software design. They provide a 'systematic' yet creative way to design and solve problems. When you're building software, you often face similar challenges, and design patterns offer a way to address these challenges in an efficient manner.
```
Lei-Tin, line 4: Perhaps you want to mention that Design Patterns can be applied to any programming languages, because they are a way of thinking instead of a specific algorithm.

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
...
+
+1. **Reusability:** Design patterns promote code reusability. Once you understand and implement a design pattern, you can apply it to different projects and scenarios. Note that it is not a piece of code for you to copy and paste, but a certain way of thinking and solving problems. 
+   - But it is not the same as _algorithm_, whose input and output satisfy a certain specifications. Design Patterns are more creative and applicable in software development, in the sense that it does not require a strict specification on its input. 
+
+2. **Maintainability:** Using design patterns results in cleaner and more maintainable code. They encapsulate specific behaviors, making it easier to modify or extend functionalities without affecting the entire codebase.
+
+3. **Scalability:** Design patterns contribute to scalable and flexible software architecture. They help in building systems that can adapt to changing requirements and scale gracefully.
+
+
+## Types of Design Patterns
```
Lei-Tin, line 16: From these 3 types, it seems like you are mentioning "Objects" a lot, maybe you want to clarify the relationship between OOP and Design Patterns? 

I believe that good OOP principles are tied to utilizing design patterns, but design patterns are not completely tied to OOP, you can also have design patterns in other programming languages.

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
...
+
+### 3. Behavioral Patterns
+Behavioral patterns define ways for objects to communicate, encapsulating patterns of communication between objects.
+
+## Simple yet Powerful Example
+We briefly introduce a **Creational Pattern** to demonstrate how powerful and flexible design patterns are. 
+
+Suppose we are developing a software where transportation tools are involved. At first, we only support cars and busses. But as our product grows, we have the need to support more types of transportation tools, like airplanes and boats, etc. We can certainly create more classes, like we did for cars and busses. But sometimes all we care about is that they are methods of transportation, i.e., they can take people from one place to another place. Let's call that behavior `move()`. 
+
+Thus, we may define a shared interface between among these classes
```
Lei-Tin, line 32: Do you want to clarify that in this case, you are trying to make the example in Python? 

But I'm not sure if it will be better if you use pseudocode here or something else.

Lei-Tin: It might be hard for people to understand what is going on with: 
`class Airplane(Vehicle):`

If they have never programmed in Python. Maybe you can kind of mention that this means "class inheritance" so that `class Airplanes` now inherits all methods defined by `class Vehicle`?

Lei-Tin: In addition to the above, maybe you can mention more on what you're doing or trying to do, so that people who are reading this can sort of "link to their own knowledge" and understand what's going on. 

E.g. you can say: 
"Here, I'm creating a class named Vehicle, which has a method that is called move()"
"Then, I am creating another class named Airplane, which inherits from the class Vehicle (which means that it will have all the methods that I have defined for class Vehicle under class Airplane). However, as Airplanes probably move in a certain way, I will "override" the definition of move that is inherited from class Vehicle, and write something else"

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
...
+```
+and 
+```python
+class Car(Vehicle):
+    def move(self):
+        ...
+```
+
+
+This way, we don't need to worry about the exact details of 'how a car/plane/bicycle' moves, but knowing that they support `move()` is good enough.
```
Lei-Tin, line 53: You didn't even create a bicycle in the code above tho
Maybe you can just remove that?

Design_Patterns.md
Lei-Tin: Also, maybe instead of "...", you can actually write some useful description so that it's easier to understand the purpose of this design pattern. 

e.g.
```
class Vehicle:
    def move(self):
        raise NotImplementedError
```

```
class Airplane(Vehicle):
    def move(self):
        # some omitted code to turn on the jet engines and make the plane fly across the skies
```

```
class Car(Vehicle):
    def move(self):
        # some omitted code to make the engines start and uses the car wheels to make the car move around
```

Lei-Tin (changes requested):
Looking good! 
Just that I think too much details are omitted in this guide, making it "maybe" hard to follow along if you're clueless on design patterns. Maybe you can try to clarify more information so that even someone that doesn't really know how to code (maybe that's too much) can follow along.

Design_Patterns.md
JY3150: I have mentioned this in my latest change

Design_Patterns.md
JY3150: I have mentioned their relationship as well.

Design_Patterns.md
JY3150: Thank you so much for your detailed comments. I have incorporated the changes you requested. Please let me know if there is anything else that you find unclear.

Teinble (changes requested):
I reviewed your wiki. Everything looks good to me besides:
1. For Why Design Patterns section, is it possible that add one or two examples that how does the example design patterns help with those good property
2. Maybe add an example for each mentioned type of design pattern.

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
+# Design Patterns for Beginners
```
Teinble, line 1: Sorry I miss-clicked the last review to approved.

Teinble (changes requested):
Check the last review comment

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Why Design Patterns?
+
+1. **Reusability:** Design patterns promote code reusability. Once you understand and implement a design pattern, you can apply it to different projects and scenarios. Note that it is not a piece of code for you to copy and paste, but a certain way of thinking and solving problems. 
+   - But it is not the same as _algorithm_, whose input and output satisfy a certain specifications. Design Patterns are more creative and applicable in software development, in the sense that it does not require a strict specification on its input. 
+
+2. **Maintainability:** Using design patterns results in cleaner and more maintainable code. They encapsulate specific behaviors, making it easier to modify or extend functionalities without affecting the entire codebase.
+
+3. **Scalability:** Design patterns contribute to scalable and flexible software architecture. They help in building systems that can adapt to changing requirements and scale gracefully.
+
```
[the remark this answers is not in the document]
JY3150: I have included another example of Observer Pattern that demonstrates maintainability and Scalability.

Design_Patterns.md
```diff
@@ -0,0 +1,74 @@
...
+```python
+class Car(Vehicle):
+    def move(self):
+        ...
+```
+
+
+This way, we don't need to worry about the exact details of 'how a car/plane/bicycle' moves, but knowing that they support `move()` is good enough. 
+
+
```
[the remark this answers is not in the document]
JY3150: I have added another example, but I think 2 examples are sufficient as I do not want to overwhelm the readers by too many examples.

Teinble (commented):
Looks great! Appreciate for your change.

PR #213, opened 2023-11-27
Create Mongoose Learning Topic

Ljime:
In this PR, I created the Mongoose Learning Topic markdown file which focuses on teaching students how to get started with Mongoose and how to use Mongoose to define Schemas that would be useful in their NodeJS projects.


Topics/Tech_Stacks/Mongoose.md
```diff
@@ -0,0 +1,179 @@
...
+	 type: Number,
+	}
+});
+```
+
+But this isn’t really exciting is it? It is just a simple basic User schema with nothing special. However, we realize that we don’t have a userId that would uniquely identify our Users right? Luckily Mongoose automatically generates an \_id field for us so we never have to worry about that unless we want a custom \_id. But in our case, we don’t need one.
+
+Now is that it? No, it’s not, let’s say we want to perform some validation on our Schema, that’s when we can use Mongoose’s built-in validation features.
+
+For example, what happens when someone creates a User without a username OR a password. That would be an invalid request right? So we can enforce that the username and password fields MUST always be present by using the “required” identifier. We can provide that the field is required and also provide a custom error message when someone creates a user without a username or a password
```
DavidZ79, line 88: Add a period at the end of the last sentence.

Topics/Tech_Stacks/Mongoose.md
```diff
@@ -0,0 +1,179 @@
...
+	  type: String,
+	  required: [true, "No Password Provided"],
+	},
+	points: {
+	  type: Number,
+	}
+});
+```
+
+Now let’s say we also don’t want people with the same username since that doesn’t really make sense to have two people named SaplingStomper27 do we? We can achieve that by enforcing the “unique” field on our username.
```
DavidZ79, line 106: lol!

DavidZ79 (changes requested):
Nice work Herman! I appreciate how concise your guide is, and how it flows very well.

Thanks for keeping the code snippets short and simple. I don't have any experience with Mongoose, but I can definitely understand what's going on. If you feel that there's anything note-worthy with a specific line of code, don't forget to add a comment or two.

It also might be helpful if you could include a bit on password hashing. As a reader, I feel that it can naturally follow your section where you touched on User schemas. We don't want SaplingStomper27 get his password stolen :) 

Also, I suggest adding some screenshots of any potential validation error the user can receive, as this can alleviate the most frustrating parts of troubleshooting for beginners.

As for the additional resources, perhaps you could include another resource that you personally used, such as a video tutorial or a blog post relating to this subject? I would also recommend adding a short excerpt about it, or why you found it to be useful. I think it would encourage a lot more people to look further into Mongoose.

Overall, i think your tutorial is a great resource for those looking for information on Mongoose, and I'll be saving it for myself. Good work!

DavidZ79 (approved):

PR #215, opened 2023-11-27
added intro to creating vue project with typescript

jianwang0722:
Added an overview of how to make a Cue 3 project that uses TypeScript.

### DIFF
--- a/Topics/Tech_Stacks/TypeScript.md
+++ b/Topics/Tech_Stacks/TypeScript.md
@@ -19,7 +19,7 @@ To play around with and experiment using TypeScript, feel free to visit the [pla
 
 ## Why use TypeScript
 
-TypeScript allows developers to catch potential errors early in the development process (at compile-time instead of runtime) and write more maintainable and scalable code. This can greatly reduce the likelihood of bugs and make code more reliable and easier to maintain, as well as improve code readability and make it easier to write reusable code. This makes TypeScript the preferred choice when building large-scale applications, especially in web development. Its syntax is similar to JavaScript, making it easy to learn and adopt for developers familiar with JavaScript.
+TypeScript allows developers to catch potential errors early in the development process (at compile-time instead of runtime) and write more maintainable and scalable code. This can greatly reduce the likelihood of bugs and make code more reliable and easier to maintain, as well as improve code readability and make it easier to write reusable code. This makes TypeScript the preferred choice when building large-scale applications, especially in web development. Its syntax is similar to JavaScript, making it easy to learn and adopt for developers familiar with JavaScript. Additionally, popular code editors such as VSCode offer additional quality-of-life functionalities such as autocompletion when writing in TypeScript, making it easier and faster to write code.
 
 To learn more about why you should consider using TypeScript over JavaScript, visit [this blog here](https://www.geeksforgeeks.org/8-reasons-why-you-should-pick-typescript-over-javascript/).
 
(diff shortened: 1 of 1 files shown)


Topics/Tech_Stacks/TypeScript.md
```diff
@@ -57,6 +57,21 @@ Angular comes with TypeScript enabled by default. In fact, Angular is built enti
 
     `ng new my-app`'
 
+### Vue
+
+To use TypeScript in your Vue 3 project using `create-vue`, which is the official Vue project scaffolding tool, follow the steps below:
+
+1. Open a terminal window, navigate to the location where the project will be created. 
+
+2. Run the command `npm create vue@latest` to initiate the project creation. (Note: if you do not have Node.js installed on your machine, follow the guide in the sections above to install it first before creating the Vue project)
```
kchprog, line 66: Consider inserting specially-formatted text (or screenshots from the terminal) at each step, so the user knows exactly what to do.

kchprog (changes requested):
Consider the commented changes; also, the existing article could use some more elaboration and direction in order for it to be more useful. Add more content and I would suggest including code snippets to provide more direction.

Topics/Tech_Stacks/TypeScript.md
```diff
@@ -57,6 +57,21 @@ Angular comes with TypeScript enabled by default. In fact, Angular is built enti
 
     `ng new my-app`'
 
+### Vue
```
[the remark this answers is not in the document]
jianwang0722: Thanks for the advice, I'll add a section at the top to summarize the steps to install node and typescript separately.

Topics/Tech_Stacks/TypeScript.md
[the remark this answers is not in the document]
jianwang0722: I'll add a section summarizing the main points in the blog post to highlight the benefits of using TypeScript, and provide some examples.

Topics/Tech_Stacks/TypeScript.md
jianwang0722: Thanks, I will include screenshots of the terminal to show the exact prompts to follow.

Topics/Tech_Stacks/TypeScript.md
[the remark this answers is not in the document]
kchprog: Sounds like a good idea.

Topics/Tech_Stacks/TypeScript.md
```diff
@@ -21,10 +21,23 @@ To play around with and experiment using TypeScript, feel free to visit the [pla
...
 
+As an example to demonstrate how TypeScript makes life easier, consider the situation where a company imports code from a third party. With JavaScript, since the IDE cannot validate JavaScript references directly, it requires a lot of care and manual testing to modify the imported code. However, with TypeScript, this issue is resolved because the IDE can automatically identify any errors that arise from incorrect references.
```
kchprog, line 24: It may be a good idea to add a reference to support this detail. I understand it to mean that we can't find errors except at runtime because of JS not having explicit typing, but it does seem a little unclear.

Topics/Tech_Stacks/TypeScript.md
```diff
@@ -67,13 +80,18 @@ To use TypeScript in your Vue 3 project using `create-vue`, which is the officia
...
 3. Follow the command line prompts to select the preferences for your project. Select `yes` for the prompt `Add TypeScript?`, and manage other settings to your preference.
 
+<img width="450" src="https://raw.githubusercontent.com/vuejs/create-vue/main/media/screenshot-cli.png"> [image source](https://github.com/vuejs/create-vue)
+
 4. Done! The Vue project is succesfully created.
 
 It is worth noting that this method is the preferred way to create Vue projects, instead of using Vue CLI, since create-vue is based on Vite. For more information, visit the create-vue Github repository [here](github.com/vuejs/create-vue).
+
+However, if you wish to set up your Vue project with Vue CLI, you can do so by following the instructions in [this video](https://youtu.be/JfI5PISLr9w?si=HEKZJlcRhWjb0F_0).
+
```
kchprog, line 90: I think it would be interesting to provide some examples of code where JS and TypeScript are compared: for  example, you could have the same simple program written in both languages and use that to show how TypeScript is easier to understand.

kchprog (changes requested):
Looks better, just have a few small adjustments to suggest.

PR #217, opened 2023-11-27
Accessibility write-up

Amyh102:
Accessibility in User Experience topic


Plopstuff:
I like the structure and information highlighted! Maybe give an example of what would be considered bad for accessibility vs good?

Plopstuff (changes requested):
I like the structure and information highlighted! Maybe give an example of what would be considered bad for accessibility vs good?

Amyh102:
> I like the structure and information highlighted! Maybe give an example of what would be considered bad for accessibility vs good?

Good point - I'll add some examples to the section below about user design

Plopstuff (approved):
I like the addition! Including resources for such examples is definitely a good idea, its usually overwhelming to figure out your own examples.

PR #226, opened 2023-11-28
a introduction to the use of redux

Brenden-Yiping-Wang:
This pull request introduces a detailed tutorial on Redux, a prominent state management library in JavaScript. Aimed at developers with a background in JavaScript. The tutorial encompasses a wide range of topics, from the basics of Redux to its core concepts, providing a thorough understanding of its functionality and usage.

### DIFF
--- a/Topics/Tech_Stacks/redux.md
+++ b/Topics/Tech_Stacks/redux.md
@@ -131,3 +131,70 @@ The state in Redux is the single source of truth for your application. It's divi
 - **UI State**: This includes state related to UI, like current active tab, loading states, user interface states like dropdowns being open or closed.
 - **App State**: This encompasses state needed globally by the whole app, like user authentication, theme settings, etc. It's the state that is relevant across different parts of your application.
 
+## Scenario Walkthrough
+
+Imagine you're building a to-do list app. The app's state, which includes the list of tasks, needs to be accessible throughout your app.
+
+### Without Redux:
+
+Each component might have its own state. Managing and syncing states across components can become complex and error-prone.
+
+### With Redux:
+
+The app's state is stored in a central place, the Redux Store. Any component can dispatch Actions to modify the state, and subscribe to state changes. This ensures consistency and predictability.
+
+Here's a basic example of how you might set up Redux in a to-do list app:
+
+#### Step 1: Install Redux
+
+```bash
+
+
+npm install redux
+```
+
+#### Step 2: Create Actions
+
+```javascript
+const ADD_TODO = 'ADD_TODO';
+
+function addTodo(text) {
+  return { type: ADD_TODO, text }
+}
+```
+
+#### Step 3: Create Reducers
+
+```javascript
+function todos(state = [], action) {
+  switch (action.type) {
+    case ADD_TODO:
+      return [...state, { text: action.text, completed: false }];
+    default:
+      return state;
+  }
+}
+```
+
+#### Step 4: Create a Store
+
+```javascript
+import { createStore } from 'redux';
+import todos from './reducers';
+
+const store = createStore(todos);
+```
+
+#### Step 5: Dispatch Actions
+
+```javascript
+store.dispatch(addTodo('Finish Redux tutorial'));
+```
+
+## Related Resources
+[Getting Started with Redux](https://redux.js.org/introduction/getting-started)
+[Redux Overview and Concepts](https://redux.js.org/tutorials/essentials/part-1-overview-concepts)
+[Redux Toolkit App Structure](https://redux.js.org/tutorials/essentials/part-2-app-structure)
+[Why Redux Toolkit is How To Use Redux Today](https://redux.js.org/introduction/why-rtk-is-redux-today)
+
+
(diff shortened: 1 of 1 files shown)


Brenden-Yiping-Wang:
> There are many state managment tools available. Why you choose introduce this one ? What are some of the strength and weakness of it?

The reason why the redux is popular nowadays is because it provides a consistent architecture with strict guidelines, which can make the code more maintainable and easier to understand for new team members. However, there is also a tradeoff. For small to medium-sized projects, the complexity and overhead that Redux introduces may not be necessary, and simpler state management solutions could suffice.

I will make a conclusion in my tutorial and push it again!

YuouLei:
> > There are many state managment tools available. Why you choose introduce this one ? What are some of the strength and weakness of it?
> 
> The reason why the redux is popular nowadays is because it provides a consistent architecture with strict guidelines, which can make the code more maintainable and easier to understand for new team members. However, there is also a tradeoff. For small to medium-sized projects, the complexity and overhead that Redux introduces may not be necessary, and simpler state management solutions could suffice.
> 
> I will make a conclusion in my tutorial and push it again!

Okay, I see. That makes sense. Thanks so much for your explanation.

Topics/Tech_Stacks/redux.md
```diff
@@ -25,6 +26,17 @@ This tutorial provides a comprehensive guide to Redux, a popular state managemen
 
 Redux is a predictable state container for JavaScript applications. It's used for managing the state of your app in a more predictable and easier-to-debug manner. Redux maintains the state of an entire application in a single immutable state tree, which can't be changed directly. When changes are made, a new object is created and merged with the state.
 
+## Why Do We Use Redux?
```
YuouLei, line 29: I see,  Redux maye be is better suited  for large scale applications. Could you explain what efforts Redux make to make the state changes maintainable and easy to understand?

Topics/Tech_Stacks/redux.md
```diff
@@ -0,0 +1,105 @@
...
+
+```bash
+npm install react-redux
+```
+
+## Core Concepts
+
+### Store
+
+The Store is the central hub where all application state is stored. It's created using Redux's `createStore` method. The Store provides several methods:
```
YuouLei, line 58: Could you provide some examples that demonstrate how these function are used? That will definitely help with understanding.

YuouLei (commented):
Generally,  some examples will defnitiely help with our understanding. Also, I have some specific question that is attached on the commits.

Topics/Tech_Stacks/redux.md
Brenden-Yiping-Wang: Redux make the state changes maintainable and easy to understand in two ways:
+ Redux maintains the entire state of your application in a single store. This centralized approach ensures that the state is consistent across the app and makes it easier to track and manage state changes.
+ In Redux, the state is immutable, meaning it cannot be changed directly. Instead, any state change produces a new state object. This immutability simplifies tracking changes over time and debugging, as each state in the application's lifecycle can be inspected and understood in isolation.

YuouLei: That makes sense a lot. So, it basically keep everything is single immutable and trackable state, which enhances the tracking of states across the whole app.

Topics/Tech_Stacks/redux.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+Actions are dispatched using the `dispatch` method of the Store.
+
+### Reducer
+
+Reducers are pure functions that take the current state and an action, and return a new state. They describe how the state changes in response to an action. Reducers should be deterministic, meaning given the same input, they should always return the same output.
+
+A reducer for a to-do application might look like this:
+
```
YuouLei, line 122: Now, I understand that the action is just a javascript object, which represents as the message sending to the store, telling what to update. And The store is like a container that has all the states. However, I don't quite get the concept of reducer. The explanation now might be hard to understand for a person without experience.

YuouLei (commented):
Action and Reducer follow up question

Topics/Tech_Stacks/redux.md
Brenden-Yiping-Wang: Definitely, I will add one example each for the dispatch, getState and dispatch.

Topics/Tech_Stacks/redux.md
Brenden-Yiping-Wang: Generally, in Redux, a reducer is a function that decides how the state of your application changes in response to an action. To put it simply, a reducer acts like a director who takes instructions (actions) and updates the state of the application accordingly. I will add a workflow explaining how to use reducer.

Topics/Tech_Stacks/redux.md
```diff
@@ -0,0 +1,133 @@
...
+- Example:
+  ```javascript
+  // Subscribing to state changes
+  const unsubscribe = store.subscribe(() => 
+    console.log('State changed:', store.getState())
+  );
+
+  // To unsubscribe from the state changes
+  unsubscribe();
+  ```
```
YuouLei, line 79: I've grasped the fundamental concepts of Redux, such as the store, action, and reducer. However, managing the state remains a challenge. Could you provide a scenario-based walkthrough, guiding a developer through a practical example of using these concepts in an app? Also, for further learning about Redux, could you recommend any online resources?

YuouLei (commented):
Follow up : Scenario walkthorugh

Topics/Tech_Stacks/redux.md
Brenden-Yiping-Wang: You are right, I will add a scenario-based walkthrough to enhance the understanding. Besides, I will also add some websites for further learning.

Brenden-Yiping-Wang:
I add the tutorial for redux, a state management library.

PR #231, opened 2023-11-28
Introduction to the Gradle Build Tool

zhaoan12:
The page will go over why we need to use build tools and why you should specifically use Gradle. Then, it will go over dependency management/dependency resolution with Gradle.


zhaoan12:
In addition to addressing all the requested changes, I added an official documentation section for further reading, and added a few comments in the code so that it is easier to understand.

PR #235, opened 2023-11-28
Fix broken links and refine props format

skatshang:
The following changes have been made to React_Native_Component.md:

1. Fixed the format of links that're not showing properly 
2. Refined the format of props sections, allowing readers to distinguish prop names and contents easily


skatshang:
thanks for checking!

PR #240, opened 2023-11-28
Added Introduction to NumPy

m-grigoriu:
Added a page to the wiki containing an introduction to the Python library NumPy, which provides a memory-efficient and flexible array data type to store values in n-dimensional collections. The introduction goes over how to include NumPy in Python projects, how arrays are created and how they work, how to use 2+ dimensional arrays, and various operations that can be conducted on/between arrays.


Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
...
+### [Introduction](#introduction-1)
+
+### [Set Up](#set-up-1)
+
+### [Array Basics](#array-basics-1)
+
+### [Matrices and Operations](#matrices-and-operations-1)
+
+### [Additional Resources](#additional-resources-1)
+
```
maxXu0713, line 14: I'm not sure why there are 1s here

Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
...
+For more detailed instructions on ways to install NumPy, there is a linked additional resource at the bottom of the page labelled 'Installing NumPy'.
+
+Finally, in order to use NumPy resources in your Python code, it is important to import the library. The general convention is to give the package the name 'np' in the import as a shorthand, however it is technically not necessary. Code blocks in this page will be written assuming that the import is done as is below:
+```py
+import numpy as np
+```
+
+## Array Basics
+The most basic object to create in NumPy is the array, which is similar to the default Python list, however it is designed to be more memory efficient, reliable, and it comes with a variety of additional methods. An array can be created with an inputted Python list:
+```py
```
maxXu0713, line 38: You already mention the import statement above but I think adding it to the coding example will make the section more modular. It would allow a reader to go directly to the example and see what they need to write to get it working right away.

Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
...
+empty = np.empty(3)
+
+# Array with zeroes/ones in each index.
+zeroArray = np.zeros(3)
+oneArray = np.ones(3)
+print(zeroArray) # '[0. 0. 0.]'
+print(oneArray) # '[1. 1. 1.]'
+```
+
+It is important to mention that arrays must contain elements of the same type. There will be an example below showcasing what happens when given a list of various types as an argumen to the array constructor.
```
maxXu0713, line 76: arugmen - spelling error

Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
+# Learning NumPy
```
maxXu0713, line 1: It may or may not be helpful to indicate in the title of the file that this is related to Python although the name includes Py so it does suggest Python. If there was a Python document I would've recommended you just add it there.

Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
...
+shortInts = np.empty(3, dtype=np.int8)
+```
+
+## Matrices and Operations
+So far, this page has only discussed one-dimensional arrays, which only represents a small part of what NumPy has to offer. In this section two-dimensional arrays (matrices) will be discussed, however arrays can be created in even higher dimensions in the same way.
+
+The matrices in NumPy act like matrices seen in algebra courses, which means that they store a two-dimensional grid of values. A matrix with 3 rows and 5 columns has shape represented by the tuple (3, 5), and an array with 5 elements has shape (5, ). There is only one shape attribute in the array's tuple due to it having only one dimension.
+
+Similar to how arrays are created, matrices can be created using a two-dimensional Python list, however it is important to note the matries must have a constant number of elements in each row (inner list), following in the same logic that an algebraic matrix cannot have missing values in a given position on the grid. Below are some examples of how to properly and improperly create matrices, an example of a 3-dimensional array, as well as some basic operations that can be carried out with them:
```
maxXu0713, line 101: matries - spelling error

Topics/Tech_Stacks/NumPy.md
```diff
@@ -0,0 +1,199 @@
...
+
+### [Array Basics](#array-basics-1)
+
+### [Matrices and Operations](#matrices-and-operations-1)
+
+### [Additional Resources](#additional-resources-1)
+
+## Introduction
+This is a short introduction to NumPy, for people looking for a basic overview of the library before doing further research into how to use it, which can be done through visiting some of the additional resources linked at the bottom of this page. Specifically, all the topics covered on this page as well as many more are discussed in both the official beginners guide and the video tutorial linked resources.
+NumPy is a Python library which provides users with new data structures which represent arrays of one or more dimensions, as well as a variety of methods for different mathematical operations which chain together effectively. It is designed for use in engineering, statistical, mathematical, as well as any other computing and/or scientific contexts where complicated operations are to be conducted on data sets of various shapes efficiently.
```
maxXu0713, line 17: What would be a use case in a more group project setting? How would this be applied in 301? The examples given seem to be more academic.

maxXu0713 (commented):
Overall, the code provides a clear and structured introduction to NumPy, covering the basics of installation and array manipulation. It looks like a useful resource for anyone trying to utilize NumPy.
I left some minor comments that include suggestions and spell checks.

m-grigoriu:
I also considered what you said about it being a Python library, but I am not sure how I could mention it naturally, especially considering that there is no Python page on the wiki (as far as I currently know). Additionally, I added '1' to the end of the table of contents endpoints as part of the convention set by other md files already present on the wiki.

Thank you for your input regarding the use case, including the import statements, and typos. I will make changes to address these comments shortly.

maxXu0713 (approved):
Looks good, the lack of a Python file does make it hard to tie together with Python. You don't have any conflict so you should be good to merge.

PR #251, opened 2023-11-28
Update CSS.md

DavidZ79:
Added a section on CSS Specificity to the existing CSS readme under Tech Stacks.

### DIFF
--- a/Topics/Tech_Stacks/CSS.md
+++ b/Topics/Tech_Stacks/CSS.md
@@ -3,6 +3,7 @@
 ## Table of contents
 ### [Introduction](#introduction-1)
 ### [CSS Cascade](#css-cascade-1)
+### [CSS Selector Specificity](#css-selector-specificity-1)
 ### [CSS Layout](#css-layout-1)
 ### [CSS Frameworks](#css-frameworks-1)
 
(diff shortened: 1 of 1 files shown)


Topics/Tech_Stacks/CSS.md
```diff
@@ -3,6 +3,7 @@
 ## Table of contents
 ### [Introduction](#introduction-1)
 ### [CSS Cascade](#css-cascade-1)
+### [CSS Selector Specificity](#css-selector-specificity-1)
```
Ljime, line 6: It may also be nice to add extra hyperlinks to the other subsections such as Specificity Hierarchy, How is specificity calculated?, Troubleshooting, Additional Resources:

Topics/Tech_Stacks/CSS.md
```diff
@@ -26,6 +27,62 @@ Because of the existence of ids and classes in selectors, often multiple CSS dec
...
 [This website](https://wattenberger.com/blog/css-cascade) gives a detailed breakdown of CSS Cascade and its tiers. 
 
+## CSS Selector Specificity:
+
+CSS specificity is the set of rules and calculations that determine which CSS selectors are executed in the CSS Cascade. Understanding this crucial in improving the fluency of a programmer's ability to write functional CSS, since this concept can save many hours of troubleshooting.
+
+#### Specificity Hierarchy
+The selector with the highest specificity is chosen to execute on the CSS Cascade. There are 3 selector categories, from most specific to least specific, shown below:
+
+1. **ID**
```
Ljime, line 37: I think !important would also be applicable here as the highest hierarchy, though it is bad practice, do you think it is worth adding?

Topics/Tech_Stacks/CSS.md
```diff
@@ -26,6 +27,62 @@ Because of the existence of ids and classes in selectors, often multiple CSS dec
...
+2. **Classes and Pseudo-classes**
+3. **Elements and Pseudo-elements**
```
Ljime, line 39: I also think the universal selector would be worth adding as well
```
* {
styles here...
}
```
The universal selector that selects all elements, would go below the Element selectors, right? I think it would be a great addition to this list!

Topics/Tech_Stacks/CSS.md
```diff
@@ -26,6 +27,62 @@ Because of the existence of ids and classes in selectors, often multiple CSS dec
...
+
+Suppose we introduce a fourth selector:
+
+`#im-the-best-id.im-another-class.im-yet-another-class { ... }  /* IDs=1, Classes=2, Elements=0 */` -> Specificity = 120
+
+This selector would have the highest specificity of all the previous examples. It has the same number of IDs as the first selector, but has 2 more classes, making it more specific.
+
+N.B. that selector combinators have no impact on the specificity calculation, and inline styles of an element have the highest specificity regardless of the categories above.
+
+#### Troubleshooting
```
Ljime, line 63: This is a really great section, I was wondering if there were any online resources to add to this such as an online calculator that would do the specificity calculations for you if you were stuck debugging? I think it would be a great resource to have for a developer!

Topics/Tech_Stacks/CSS.md
```diff
@@ -26,6 +27,62 @@ Because of the existence of ids and classes in selectors, often multiple CSS dec
...
+
+If you believe that your CSS is behaving in unexpected ways due to selector specificities, here are some steps you can take (in no particular order) to look into what is causing it.
+
+1. Use Browser Developer Tools
+
+Every browser has developer tools that you can take advantage of when debugging your CSS. They allow you to see which selectors have been executed, and which have not. In Google Chrome, Right-Click -> Inspect Element or F12 to access Chrome Dev Tools. Under the styles tab, you can see which styles have been crossed out, meaning that they have not been executed.
```
Ljime, line 69: Do you think you could add images or link a tutorial with this guide? I think a visual guide may help those who haven't used developer tools before!

Ljime (changes requested):
Hi David, 

Overall, this is a really nice update to the CSS documentation and a must-read for every CSS developer! I have suggested some changes and additions that I think would make this more a thorough guide.

Great work overall!

Topics/Tech_Stacks/CSS.md
DavidZ79: I think I disagree with adding hyperlinks to the table of contents, since it is meant to be a high-level list of concepts, and to keep consistent with other topics, which aren't as extensively written about. Should there be more contributions in the future, I'll be sure to add some subpoints to the table.

Topics/Tech_Stacks/CSS.md
DavidZ79: Good catch! Could be just me, but I'm not sure that everyone who would read this would already know about !important. I'll be sure to add it.

Topics/Tech_Stacks/CSS.md
DavidZ79: I actually didn't know about that, thanks

Topics/Tech_Stacks/CSS.md
DavidZ79: I have limited experience using a specificity calculator, but I can understand why it can be tremendously useful in troubleshooting css. I'll add it, thanks

Topics/Tech_Stacks/CSS.md
DavidZ79: You're right! I added an image along with a tip to check/uncheck certain styles.

DavidZ79:
Thanks for all your comments! I made a few changes here and there to reflect them.

Ljime (approved):
Hi David.

Thank you for the changes!
Everything looks good to me now!

Good job!

PR #253, opened 2023-11-28
First draft of [Asking For Help] Section

AndrewNQN:
Added a section on how to ask for help. I think that this will be useful for the students, like myself, that had little to no experience of working with code in a large group. I spent the first few weeks struggling with getting started so I think that this would help.

### DIFF
--- a/Topics/Teamwork.md
+++ b/Topics/Teamwork.md
@@ -63,6 +63,7 @@ Letting your emotions overwhelm you can also hinder effective communication in t
 
 ## Asking For Help
 
+### With a Coding Problem
 A new environment produces new challenges. It may be the case that you are lost and are unsure of how to proceed. Here is a list of things you can do, as an individual and as a team member, to get yourself back on track.
 
 1) Attempt to fix the situation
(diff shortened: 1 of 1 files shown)


manya-mittal (changes requested):
This is great, asking for help when you're stuck can be difficult so this will be very useful! You mentioned in your description that you found it difficult to start in your group's project. Could you expand on this in the PR? I think it could be helpful for people to know how to went about figuring out where to begin. Also, your current version focuses on not knowing how to code a particular functionality. Maybe you could include another section that outlines how to ask for help when it is not coding related?

AndrewNQN:
Thank you Jessica.

1) I decided not to make any changes regarding this point as I believe that this topic is better suited elsewhere. My advice is solely focused on the process of asking for help, rather than keeping code simple and clean to avoid misunderstandings that will lead to asking for help.

2) I added a summary and an extra subsection outlining a few things you should avoid doing.

Please look over the new editions and tell me what you think.

SiqiZhang6 (approved):
I see, thank you Andrew, this looks great :)

manya-mittal (approved):
Thanks for making the changes andrew, it looks good to me now!

PR #256, opened 2023-11-28
Added content to the Giving and Receiving Feedback section

jchang0112:
Provides a feedback giving process that allows recipients to understand and accept feedback more easily, while feeling more positive emotions.


Eshchock1 (changes requested):
Looks pretty good, like the content so far but here are some suggested changes:

- Consider changing the heading to just "Giving Feedback"
- Include specific examples for each step of delivering feedback
- Describe the impact of feedback more effectively

jchang0112:
Could you clarify what exactly you mean by describing the impact of feedback more effectively? I have made the other changes asked, but I am not entirely sure how I could go about describing the impact, should I do something along the lines of comparing well delivered feedback to not well delivered feedback?

Eshchock1:
> Could you clarify what exactly you mean by describing the impact of feedback more effectively? I have made the other changes asked, but I am not entirely sure how I could go about describing the impact, should I do something along the lines of comparing well delivered feedback to not well delivered feedback?

I've taken a look at your new changes, and they look good. When it comes to describing the impact of feedback I agree with comparing well-delivered feedback to not well-delivered feedback. Here are some suggestions in specific:
- Discuss in more detail why feedback is necessary and beneficial
- Consider discussing the difference between constructive/destructive feedback

jchang0112:
I have made the changes you suggested, including an example of destructive feedback.

Eshchock1 (approved):
Nice changes, looks good to me!

PR #262, opened 2023-11-28
create test driven development markdown

eric-pham:


wang-henry:
I think you should include a coding example of what it would look like to write a test that fails at first but matches the requirements. It would be easier to conceptualize with an example.

wang-henry:
Also remember to update your topic in `Development_process.md` before merging.

eric-pham:
> I think you should include a coding example of what it would look like to write a test that fails at first but matches the requirements. It would be easier to conceptualize with an example.

Oh that's a good point, I'll include actual examples

larryworm1127:
Are there any additional resources that you recommend if people want to know more about this concept?

larryworm1127:
I think you should also explain the differences between an E2E test and a unit test, and how they relate to TDD

wang-henry:
> I think you should also explain the differences between an E2E test and a unit test, and how they relate to TDD

I think that unit tests and E2E tests are a little out of scope of this topic, that seems like it would be a bigger topic and could probably even be its own.

eric-pham:
Yeah I agree with Henry, that topic is a little more specific into what an E2E test does vs what a unit test does but I will mention something about that and just quickly talk about the differences in testing

larryworm1127 (approved):

PR #266, opened 2023-11-29
Merge selenium topic to main branch

jfitzgerald1126:
This PR adds a topic under the tech stacks section. This new topic links to a guide on web scraping and automation using Selenium in Python.


Topics/Tech_Stacks/selenium.md
```diff
@@ -0,0 +1,94 @@
...
+    Summary: 
+    Home-page: https://www.selenium.dev
+    Author: 
+    Author-email: 
+    License: Apache 2.0
+    Location: /Users/john/CSC301/venv/lib/python3.11/site-packages
+    Requires: certifi, trio, trio-websocket, urllib3
+    Required-by: 
+    ```
+2. **Install Chrome WebDriver**
```
chenbriann, line 57: This part of the installation guide could use more detail. How could I find out what version of Chrome I am using? What website do I download the Chromedriver from? Are there different kinds of Chromedrivers, and if so, what are the differences between them?

Topics/Tech_Stacks/selenium.md
```diff
@@ -0,0 +1,94 @@
...
+display.start()
+```
+
+Then, after the interaction runs, close your virtual display.
+```python
+display.close()
+```
+Typically, you would use a _try, except, finally_ statement to execute the code, and you would instantiate the display at the top of the `try` block, and close it at the end of the `finally `block.
+
+## Automation
```
chenbriann, line 81: I think it would be useful for readers to have a simple example script that demonstrates Selenium automation. That way, they have a foundation for how to start writing scripts using Selenium.

chenbriann (changes requested):
Overall I find this guide informative and easy to follow. For the most part I think a reader can get started with using Selenium fairly quickly using this guide. I commented in the "Getting Started" and "Automation" sections a couple of improvements to be made that will make it easier for the reader to install Selenium and start writing automated scripts. Great work!

Topics/Tech_Stacks/selenium.md
jfitzgerald1126: Good point, I will add this information for the reader's reference.

Topics/Tech_Stacks/selenium.md
jfitzgerald1126: Right, I agree that an example makes sense.

chenbriann:
Thanks for considering my suggestions! I can see that you have added clear instructions for installing Chromedriver and an example of using automation. One last thing I forgot to mention is perhaps adding a table of contents. Other than that, I think this should be ready to merge!

jfitzgerald1126:
> Thanks for considering my suggestions! I can see that you have added clear instructions for installing Chromedriver and an example of using automation. One last thing I forgot to mention is perhaps adding a table of contents. Other than that, I think this should be ready to merge!

Thank you for the feedback!

PR #268, opened 2023-11-29
chanju45 - Managing_Environment_Variables_and_Secrets

snowtheghost:


NeuroEnvision:
Hi Justin, 
Great work! I found this guide very helpful for someone who wants to learn about managing environmental variables and secrets in Java Spring Boot applications. You explain things in such a concise and easy-to-understand manner. 

I have a few small changes I'd like you to consider:

1. Do you want to add some in-line comments to the code block in section "Securing Secrets"? 
Specifically, you can highlight the statement on line 52 is the most relevant line of code here and explain briefly what the function `generateToken` is doing just in case someone unfamiliar with JWT token is distracted by the function. 
2. Do you want to specify when a developer can load the env var `JWT_SECRET` using the terminal command `export` (line 86)? 
For example, do they open another terminal and load this env var right after launching the Spring Boot application? 
3. Do you want to provide links for further readings on using encryption and decryption libraries (line 97)?
4. Do you want to mention that after we indicate the active environment is the deployment environment, for the environment variables present in both `application.properties` and `application-deployment.properties`, the specifications in `application-deployment.properties` have higher priority and overwrite those in `application.properties`. But for the env var only specified by `application.properties`, they will remain effective.

snowtheghost:
> Hi Justin, Great work! I found this guide very helpful for someone who wants to learn about managing environmental variables and secrets in Java Spring Boot applications. You explain things in such a concise and easy-to-understand manner.
> 
> I have a few small changes I'd like you to consider:
> 
>     1. Do you want to add some in-line comments to the code block in section "Securing Secrets"?
>        Specifically, you can highlight the statement on line 52 is the most relevant line of code here and explain briefly what the function `generateToken` is doing just in case someone unfamiliar with JWT token is distracted by the function.
> 
>     2. Do you want to specify when a developer can load the env var `JWT_SECRET` using the terminal command `export` (line 86)?
>        For example, do they open another terminal and load this env var right after launching the Spring Boot application?
> 
>     3. Do you want to provide links for further readings on using encryption and decryption libraries (line 97)?
> 
>     4. Do you want to mention that after we indicate the active environment is the deployment environment, for the environment variables present in both `application.properties` and `application-deployment.properties`, the specifications in `application-deployment.properties` have higher priority and overwrite those in `application.properties`. But for the env var only specified by `application.properties`, they will remain effective.

Hi Anqi, thank you for taking the time to review my work :)

As for point 1, this may be a good addition, though since the purpose of this text is only to walk through environment variables, I may keep the elaboration short.

I will definitely take your advice for points 2 and 3. However, for point 4, I'm not actually entirely sure that this is how Spring handles the property files. I will do some research and add clarification to this point. 

Thanks again Anqi! Good luck with your submission :)

snowtheghost:
> > Hi Justin, Great work! I found this guide very helpful for someone who wants to learn about managing environmental variables and secrets in Java Spring Boot applications. You explain things in such a concise and easy-to-understand manner.
> > I have a few small changes I'd like you to consider:
> > ```
> > 1. Do you want to add some in-line comments to the code block in section "Securing Secrets"?
> >    Specifically, you can highlight the statement on line 52 is the most relevant line of code here and explain briefly what the function `generateToken` is doing just in case someone unfamiliar with JWT token is distracted by the function.
> > 
> > 2. Do you want to specify when a developer can load the env var `JWT_SECRET` using the terminal command `export` (line 86)?
> >    For example, do they open another terminal and load this env var right after launching the Spring Boot application?
> > 
> > 3. Do you want to provide links for further readings on using encryption and decryption libraries (line 97)?
> > 
> > 4. Do you want to mention that after we indicate the active environment is the deployment environment, for the environment variables present in both `application.properties` and `application-deployment.properties`, the specifications in `application-deployment.properties` have higher priority and overwrite those in `application.properties`. But for the env var only specified by `application.properties`, they will remain effective.
> > ```
> 
> Hi Anqi, thank you for taking the time to review my work :)
> 
> As for point 1, this may be a good addition, though since the purpose of this text is only to walk through environment variables, I may keep the elaboration short.
> 
> I will definitely take your advice for points 2 and 3. However, for point 4, I'm not actually entirely sure that this is how Spring handles the property files. I will do some research and add clarification to this point.
> 
> Thanks again Anqi! Good luck with your submission :)

I did more reading and Anqi is indeed correct. I have added a further reading link on this topic as it is not in the scope of my topic.

NeuroEnvision:
> > Hi Justin, I reviewed your changes and they look good! Now I understand that instead of hardcoding env var/secrets into source code or property files, we can use ${env_var} as a sort of placeholder in property files and feed the actual values at runtime. I still have one question: how does a developer load multiple env vars all at once from the command line? Do they write a script to automate this process? If yes, where do they store the script? But I'm not sure whether it would be relevant to briefly answer this question in this guide.
> 
> Hi Anqi, thank you for looking over it. That's a very good point. A script does work, but for cloud services such as AWS, GCP, and Render like we used, there's an interface to load them in. If we're on a production server, we might hold those keys in a yaml or env file which can be parsed, though I'm not sure what common practice is when it comes to using a self-hosted server.

Thanks for the clarification!

snowtheghost:
> > > Hi Justin, I reviewed your changes and they look good! Now I understand that instead of hardcoding env var/secrets into source code or property files, we can use ${env_var} as a sort of placeholder in property files and feed the actual values at runtime. I still have one question: how does a developer load multiple env vars all at once from the command line? Do they write a script to automate this process? If yes, where do they store the script? But I'm not sure whether it would be relevant to briefly answer this question in this guide.
> > 
> > 
> > Hi Anqi, thank you for looking over it. That's a very good point. A script does work, but for cloud services such as AWS, GCP, and Render like we used, there's an interface to load them in. If we're on a production server, we might hold those keys in a yaml or env file which can be parsed, though I'm not sure what common practice is when it comes to using a self-hosted server.
> 
> Thanks for the clarification!

Thank you! I've updated the PR with a note about loading in environment variables

NeuroEnvision (commented):
I think your PR is ready to merge. Good job!

PR #269, opened 2023-11-29
Create React_Bootstrap_modals.md

maxXu0713:
Added a page to the wiki for the modal component of React-Bootstrap, which allows for a pop-up that can be used to notify the user or extract information. The introduction goes over the setup, how to connect the display and dismissal to a button, and how to extract information from the user.

### DIFF
--- a/Topics/Tech_Stacks/React_Bootstrap_modals.md
+++ b/Topics/Tech_Stacks/React_Bootstrap_modals.md
@@ -124,7 +124,7 @@ The full Modal can be displayed statically if state isn't involved and the "show
 ```
 
 ## Extracting Data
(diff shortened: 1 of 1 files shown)


maxXu0713:
Thanks for your feedback regarding my table of contents, phrasing, variant explanation, and typos. I have adjusted everything accordingly, please have a look.

In terms of a relevant use case, I said the modal could "_notify the user or obtain user information._". I think this is a very common and useful feature that is needed for most 301 projects.

m-grigoriu (approved):
Your document looks a lot better now, and I understand what you mean about the CSC301 applications. Since there do not seem to be any conflicts, I would say that you are ready to merge your PR.

PR #284, opened 2023-11-29
Pythonic Iteration Patterns

shape-warrior-t:
Decided to write about idiomatic Python -- more specifically, some indispensable convenience functions and general idioms surrounding iteration that I know very well but evidently not everyone does. Not the more abstract concepts -- the simple, (relatively) easy-to-understand stuff that anyone can just use upon becoming aware of it.

Any thoughts? Scope, focus, clarity of exposition, length, errors, etc?


Topics/Tech_Stacks/Pythonic-Iteration-Patterns.md
```diff
@@ -0,0 +1,238 @@
...
+# Write:
+for i, (a, b, c) in enumerate(zip(lst_a, lst_b, lst_c)):
+    print(f'{i}: {a} & {b} & {c}')  # prints '0: 1 & 2 & 3', then '1: 5 & 6 & 5', then '2: 9 & 5 & 8'
+```
+Note that this is only equivalent if all the arguments to `zip` have the same number of items to iterate over.
+By default, iteration cuts off when any of the arguments to `zip` run out of items to iterate over.
+`zip` also supports an optional `strict` keyword argument to raise an error
+if the given iterables do not have the same number of items to iterate over.
+Also see [`zip_longest` from the `itertools` module](
+https://docs.python.org/3/library/itertools.html#itertools.zip_longest).
```
johnnyjly, line 127: This example is duplicate because you've provided the examples for enumerate above and this is irrelevant to zip. If you want to show that the return object of zip (i.e., iterator of tuples) can also be applied to enumerate, it's better to explicilty show the complete input/return types of these built-in functions.
Also, for the 'default' behavior and 'strict' keyword, it's better to provide 2 more examples. (I find the examples in the python documentation easy to follow. https://docs.python.org/3/library/functions.html?highlight=enumerate#zip)

Topics/Tech_Stacks/Pythonic-Iteration-Patterns.md
```diff
@@ -0,0 +1,238 @@
...
+For example:
+```python
+lst = [3, 1, 4, 1, 5, 9]
+
+# Instead of:
+sum([x * 2 for x in lst])
+
+# Simply write:
+sum(x * 2 for x in lst)
+```
```
johnnyjly, line 224: The link above is good (enables the reader to have a good understanding of the generator), however, it'd be better to add more details in this example.
Compared to the examples in the link, the example here seems to be confusing.

Topics/Tech_Stacks/Pythonic-Iteration-Patterns.md
```diff
@@ -0,0 +1,238 @@
...
+First off, instead of writing `[x for x in iterable]` or `{x for x in iterable}`,
+you can simply write `list(iterable)` and `set(iterable)`. Similar functionality also exists for `tuple` and `dict`.
+
+
+### Grabbag of convenient functions
+
+Python comes with functions to compute the sum, product, maximum, or minimum of any iterable:
+`sum`, `prod` from the `math` module, `max`, and `min`.
+If you want to check if a criterion is satisfied for every item in an iterable, use the `all` function;
+for at least one item in an iterable, the `any` function.
```
johnnyjly, line 184: Would be better to provide some simple examples here.

johnnyjly (changes requested):
Overall is perfect.
It'd be better if you could provide more details on those built-in functions (input parameters, return types, and etc.).
For the rest, check the comments below.

shape-warrior-t:
Revised based on your comments. Instead of writing out the parameter and return types myself, I decided to just link to the Python docs for the major functions showcased. Added the requested examples, and reorganized the iterator-combining example into its own section alongside a modified and re-explained generator comprehension example. Please let me know if there are any more issues!

johnnyjly (changes requested):
After reading your changes and the whole file again, I think these python patterns are indeed valuable resources for anyone looking to improve their iteration codes in python.

However, I have just a few more small requests (as shown below) before I approve. Also, you may need to update the corresponding topics in the README.md and Tech_Stacks.md.

Topics/Tech_Stacks/Pythonic-Iteration-Patterns.md
```diff
@@ -0,0 +1,284 @@
...
+## Other tools to look out `for`
+
+As a rule of thumb, if you find yourself writing code that looks something like this:
+```python
+for index in indices_of_a_collection:
+    do_something_with(index, collection[index])
+```
+...there's probably a more idiomatic way to go about things.
+
+Here are a few of those potential ways to go about things:
```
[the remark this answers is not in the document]
shape-warrior-t: Decided to reword things slightly here. Still think there should be some kind of additional lead-in to things, but now it's worded in a less redundant-sounding way.

johnnyjly (approved):
Great! 
Everything looks perfect and there are no conflicts with the main branch.
I think you're good to merge this Pull Request.

PR #295, opened 2023-11-29
add example for SOLID

Christina0313:
Updated SOLID Principle description and write the advantage separately. Add examples for the use of it


yumozi (changes requested):
This seems like a decent introduction to SOLID principles. The examples of each principle in a restaurant scenario should be especially helpful in understanding the SOLID principles. 

However, a few minor issues:
- The definition of each principle could be more concise and clear. For example, when describing SRP, "A class should only have one cause to change. and "...which implies that there ought to be just one motivation for change" seem redundant. 
- The reason that some principles offer their advantage is not clear. For example, why exactly does OCP promote stability and reduce bugs? This advantage can be said for any SOLID principle, so it would be beneficial to outline a more explicit reason in relation to OCP.
- Providing some code snippets in the restaurant example would be very beneficial, but not required.

Christina0313:
> This seems like a decent introduction to SOLID principles. The examples of each principle in a restaurant scenario should be especially helpful in understanding the SOLID principles.
> 
> However, a few minor issues:
> 
> * The definition of each principle could be more concise and clear. For example, when describing SRP, "A class should only have one cause to change. and "...which implies that there ought to be just one motivation for change" seem redundant.
> * The reason that some principles offer their advantage is not clear. For example, why exactly does OCP promote stability and reduce bugs? This advantage can be said for any SOLID principle, so it would be beneficial to outline a more explicit reason in relation to OCP.
> * Providing some code snippets in the restaurant example would be very beneficial, but not required.

Great suggestion! I will make changes according to the advice. Some code example of SOLID is included in the video. I will try to take a look at the code for restaurant but some of them are more easy to explain use just word.

yumozi (approved):
Looks much better! Each principle is explained clearly and the reasons behind the advantages are also clear.

Christina0313:
> Looks much better! Each principle is explained clearly and the reasons behind the advantages are also clear.

Thank you for the feedback!

PR #296, opened 2023-11-29
Intro to REST

MacroSony:


Yuhan-Ut (commented):
Overall good wiki page for RESTful API standards information! Considering it's a standard of making API and make to use idea, it would be great for you to provide some example code! Thus beginner will know what's going on!
But still, you are going very good job hear! 

By the way, I think there is a conflict with main, that you can resolve in one minute, please take some time to do it.

Yuhan-Ut (approved):
good work!

PR #299, opened 2023-11-29
IPC with fork and pipe in C

JacqJingyuH:


MacroSony (commented):
Great introduction and a lot of useful details! 

Due to the complexity of IPC, I think it would be helpful to include some diagrams to better demonstrate how parents and children processes communicate.

Also, you forgot to link to the index page.

Yuhan-Ut (commented):
It's a great introduction about IPC with fork and pipe in C !
Because of the usefulness of IPC, It would be great for you if you can add more useful code example, so people can easily see how to use it and be confident to use it in their own code! It's never bad to have more example!

Also, I noticed that there are other IPC approaches, it would be such good idea to mention some other IPC approaches, so this wiki page will be more complete and useful!

MacroSony (approved):
Great work!

PR #305, opened 2023-11-29
Adding a page for "Debugging in Python"

manya-mittal:
This PR adds a new page to the software engineering wiki about debugging in python. While working on my team''s project, I found that the majority of my time was spent on debugging. At first, I was trying to use a lot of print statements to debug, but this quickly got messy. Using the in built python debugger "pdb" was very useful for me to learn in order to streamline the debugging process. I believe that adding this page will help future students to learn better debugging practices.


AndrewNQN (changes requested):
Nice. I've been using breakpoints for a while and they really help with isolating the exact point that the error occurs. I do think that you could expand on this topic though. There are many other debugging tools out there for Python. I see that you mentioned PyCharm and Sentry as additional tools. I think that you should expand on these subtopics and add more detail about how they work and how to use them. Additionally, if it makes sense to do so, it may be good to add when you would want to use one tool instead of another.

manya-mittal:
Thanks Andrew. I've expanded on the additional tools section now. I've added explanations for each tool, as well as when they may be picked over pdb. I have also added a few links for people to read about them in further detail. Please let me know if you think I should make any other changes

SiqiZhang6 (commented):
This looks great, I just have a couple clarification questions and thoughts:
1) you mentioned pdb is built in python, thus I'm assuming we don't need to install anything? (as some import statements also require installation of packets)
2) do we just run the program in order to start debugging for pdb? It would be nice to clarify that though that could be intuitive. If so, maybe one advantage of pdb is that it's compatible with many IDEs, and con might be it would be harder to delete all breakpoint statements once debugging is finished
3) it would be more straight forward if you have a table of pro and cons for pycharm, Sentry, and pdb

manya-mittal:
Thanks for your comments @AndrewNQN and @SiqiZhang6!

I have fix the typo, and added some clarifications about how to run pdb. I have also added a table comparing the pros and cons of the tools mentioned. Please let me know if you think I should change anything else!

AndrewNQN (approved):
Good addition at the end. I have no more issues with this PR.

SiqiZhang6 (approved):
looks amazing ty for making the changes

PR #313, opened 2023-11-30
Create Regex.md

kellythle:
Created Regex.md with an introduction, basic regular expressions, and extra resources

### DIFF
--- a/Topics/Tech_Stacks/Regex.md
+++ b/Topics/Tech_Stacks/Regex.md
@@ -1,0 +1,67 @@
+# Regular Expressions (Regex)
+
+## Table of contents
+### [Introduction](#introduction)
+### [Basic Regular Expressions](#basic-regular-expressions)
+### [Extra Resources](#extra-resources)
+
+## Introduction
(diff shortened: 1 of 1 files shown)


kellythle:
Thank you for your suggestions. I have:
- Added language support sections
- Added examples for each regular expression pattern
Let me know if there is anything else I should add or change!

JiwooJiyof:
The additional changes make the documentation a lot more clear and easier to understand. Another point I want to suggest is adding a visual diagram of how Regex is interpreted. Also, the documentation could become a lot more practical if there is a common use case examples. For example, you could add in email validation or website URL matching.

kellythle:
Thank you again for your suggestions. I feel that students can definitely benefit from common case uses and understanding where regex can be applied. I have:
- Added a section for common case uses with examples and explained why some of these may not always be practical
- Added an additional resource for visualizing an inputted regex through railroad diagrams

Please let me know if there are other parts I can add or change!

JiwooJiyof:
The addition of greedy vs. lazy quantifiers look very well-explained. Currently, the table of contents does not work properly for me. After that has been resolved, the branch should be good to merge.

kellythle:
Great catch! Thank you. I have fixed the table of contents and it seems to work properly now. Please feel free to merge the branch if everything works!

JiwooJiyof:
Everything looks great. I merged your branch. Great work!

PR #316, opened 2023-11-30
Reactstrap introduction

rileymccarron:
Created the Reactstrap.md file in the Tech_Stacks directory. The file contains an introduction to the Reactstrap library for React and how to utilize it for creating styled components without the use of CSS for frontend development.

The sections included are: an introduction to the tutorial and what you can expect to get out of it along with the prior knowledge you should have, an introduction to what Reactstrap is and why it is used, how to install and setup Reactstrap in a React project, the basics of utilizing Reactstrap components as well as an example and some information on external sources.


PyroPoro (commented):
Hi Riley,

The page looks good to me. I don't see any errors.  
The page is organized neatly and flows from start to end.  
I like how the reader is told what they should expect in the preface.  
I would suggest including a section outlining why one may choose to use Reactstrap over other libraries or frameworks (if there exists similar ones that achieve similar things, I'm not familiar enough with web dev to know).

PyroPoro:
One thing i've noticed is that the external links you have replace the current webpage, maybe you could consider changing those to open in a new tab

rileymccarron:
I have changed the table of contents links and it seems that github does not support this behaviour, but it is in proper format now and should work once posted on the website

PyroPoro:
Everything looks good. I think its ready to merge

rileymccarron:
I actually just found a type on the last line, fixing that now

PR #319, opened 2023-11-30
created count_vectorizer.md

nomiko0626:


kaavyakalani26:
I think you can make a table of contents for you page making it easier for a user to follow the page.

kaavyakalani26 (changes requested):
I think you can make a table of contents for you page making it easier for a user to follow the page.

nomiko0626 (commented):
You're correct. I'll include a table of contents.

kaavyakalani26:
The Contents table looks good. Though there is a way to link the contents to the topics in md which you can explore!

Also noticed your renaming the file and moving to the right folder, something I was going to comment so that is great!

kaavyakalani26:
Additionally, maybe you can add some more usages?

nomiko0626:
Thanks for the feedback! I've made all the changes you've requested.

kaavyakalani26 (changes requested):
It looks good! Just one last small thing, for the anchors it should be (#how-it-works) and (#additional-resources)

nomiko0626:
Fixed it.

kaavyakalani26 (approved):
Looks all good now! Thank you.

PR #321, opened 2023-11-30
Create SQL-creating_database.md

kaavyakalani26:
The first draft of my markdown file. My file introduces schemas in SQL and steps and tips to create a basic database structure.

### DIFF
--- a/Topics/Tech_Stacks/SQL-creating_database.md
+++ b/Topics/Tech_Stacks/SQL-creating_database.md
@@ -47,20 +47,22 @@ CREATE TABLE <tablename> (
 <br>
 <column1_name column1_datatype other_attributes,
 <br>
-column2_name column2_datatype other_attributes,...);
+column2_name column2_datatype other_attributes,
+<br>
+...);
 
-Example :
-CREATE SCHEMA University
-CREATE TABLE Students(
-    utorid varchar(8),
-    fname varchar(15),
-    lname varchar(15)
+Example : <br>
+CREATE SCHEMA University<br>
+CREATE TABLE Students(<br>
+    utorid varchar(8),<br>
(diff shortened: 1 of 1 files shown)


keviny15:
I think each subtopic should be added to the table of contents just for more clarity and organization

kaavyakalani26:
Thats a good suggestion! I will do that.

keviny15:
I also think that some examples would be helpful for readers. Maybe in the How to create tables section it would be helpful for me to see what the table would look like after writing the command!

Also for the redundancy section, I think you can go more in depth about common ways to fix it similar to what we see with design patterns and code smells if possible.

kaavyakalani26:
Thank you for the suggestions!
For the example, do you mean an example of the query needed to create a table? Or that along with the table which would be an empty table with columns?  

Also for the redundancy, I did add links to fix it but I think you are right. I will add a bit of information talking about ways to fix it

keviny15:
> Thank you for the suggestions! For the example, do you mean an example of the query needed to create a table? Or that along with the table which would be an empty table with columns?
> 
> Also for the redundancy, I did add links to fix it but I think you are right. I will add a bit of information talking about ways to fix it

Yeah I meant the resulting table, maybe even some example values to fill the empty table would be helpful too!

keviny15:
Maybe rephrase 4. Avoid designing your schema in such a way that there are attributes that can be null.
Into : Construct your schema with the intention of minimizing the presence of nullable attributes.

Forgot colons on lines 29, 33, 40.

keviny15:
Thank you for making the changes, looks great to me and I learned a lot!

kaavyakalani26:
Thank you for your comments! I will merge the branch now

PR #325, opened 2023-11-30
learning about QRcode

Yuhan-Ut:


MacroSony (commented):
A great introduction to QR code! I think the error correction level is an interesting part that didn't get a lot of explanation, it would be great if you could talk more about how error correction works and how different error correction levels affect the QR code's pattern.

PR #339, opened 2023-11-30
Initial Commit for Learning_Spring_Data_JPA.md

NeuroEnvision:
This article introduces readers to using Spring Data JPA for relational databases with a hands-on approach.


Topics/Tech_Stacks/Learning_Spring_Data_JPA.md
snowtheghost, line 1: Great topic! Easy to read structure and covers a lot. Some points I've thought of:

- Under Step 2, it would be worth mentioning that this is just a running example and not necessary to actually get things working. Although it's quite clear to me, it may not be to someone who is completely new to JPA. 
- Your running example is great and explores all the different operations you could make. However, there's no introduction to these operations without the example. Perhaps it would be nice to include a summary of what operations JPA supports before giving those examples.
- Your note about column names being automatically parsed and "translated" into SQL naming convention is great. It would be good to also mention that you can override this using the @Column.name attribute. See more [here](http://www.java2s.com/Tutorials/Java/JPA/0120__JPA_Column_Name.htm)
- You mention JPA creates a table for us. This is true, but what happens when we want data to persist and update the table rather than creating a new one every time? This could be worth mentioning, especially the necessary properties that need to be present. For example in hibernate we have `spring.jpa.hibernate.ddl-auto = create`

Overall a great writeup. I'm looking forward to reviewing your changes

NeuroEnvision: Hi Justin,

I made some changes based on your feedback. Can you take a look?
Specifically, I 
1. Added a summary of the functionality and benefits of Spring Data JPA. 
2. Added a link for configuring with persistent database
3. Clarified using @Column.name to override field name

snowtheghost: Absolutely. I think you nailed it. The summary and explaining that using H2 as a running example made things a lot clearer. One last thing I feel as though would wrap up the writeup very nicely is an additional readings section with links to further reading. What do you think?

NeuroEnvision: Great point! Let me include some resources I found helpful.

NeuroEnvision: > Absolutely. I think you nailed it. The summary and explaining that using H2 as a running example made things a lot clearer. One last thing I feel as though would wrap up the writeup very nicely is an additional readings section with links to further reading. What do you think?

Hi Justin, 
I added a further reading section. What do you think?

PR #355, opened 2023-11-30
Add Ray wiki

jimschenchen:
This is a introduction of one primitive in Ray Core, statless tasks, and how you can use it to distribute and execute your functions on multiple machines.

### DIFF
--- a/Topics/Tech_Stacks/Ray.md
+++ b/Topics/Tech_Stacks/Ray.md
@@ -1,0 +1,127 @@
+# Using Python Ray Tasks to distribute your functions on multiple machines
+
+
+## Table of Contents
+- [Prerequisites](#prerequisites)
+- [Introduction](#introduction)
+- [Introduction to Ray](#introduction-to-ray)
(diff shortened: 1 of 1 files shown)


Galaxy0229:
Looking at your intro, I think Python Ray would be useful, but I'm not quite sure what specific scenario and project it would use it in, can you elaborate on this aspect?

jimschenchen:
> Looking at your intro, I think Python Ray would be useful, but I'm not quite sure what specific scenario and project it would use it in, can you elaborate on this aspect?

Ok thanks for your suggestion. I will add use scenarios.

Galaxy0229:
> > Looking at your intro, I think Python Ray would be useful, but I'm not quite sure what specific scenario and project it would use it in, can you elaborate on this aspect?
> 
> I've added Scenarios for use section. Is this comprehensive?

Yeah that is great for me. Thanks for the addition :)

Galaxy0229 (approved):
Chenyue did nice job :)

PR #304, opened 2023-11-29
Add Kanban Introduction and Methodology

Tommy-Gong:
A Markdown file focused on Kanban. 
1. Basic Introduction to Kanban: The document starts with a foundational overview of Kanban, outlining its principles and significance in a software development context. This section provides a solid base for newcomers to understand the core concepts of Kanban.

2. Building a Kanban System: I have detailed a step-by-step guide on establishing a Kanban system. This guide is designed to be practical and easy to follow, helping team members implement Kanban in their workflows effectively.

PR #369, opened 2023-12-01
re-push the kanban.md

Tommy-Gong:
The original KanBan has some problems. All comments and reviews are in "Add Kanban Introduction and Methodology #304". This is just for resubmission.


Tommy-Gong:
The original KanBan has some problems. All comments and reviews are in "Add Kanban Introduction and Methodology #304". This is just for resubmission.

PR #376, opened 2023-12-01
Setup Guide on Vue Components and Other Tips!

Fight-Fighter:
This is a guide on Vue Components. I liked working with Vue, but I had quite a bit of trouble at the start due to some confusing documentation and deprecated code that I was working with. I hope to give users all they need to know about Vue components, as well as other tips that will help them learn more.


Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+
+## Table of Contents
+
+#### [Introduction](#Introduction)
+#### [What is a Vue Component? ](#what)
+#### [Topics](#thing)
+#### [Topics again](#something)
+#### [Other Tips](#other)
+#### [Conclusion](#conclude)
+
```
Firebrandt, line 11: This is smart to have. Good formatting here.

Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+If you have no knowledge of Vue or how to set up a Vue project, it is imperative that you follow [Vue's Quick Start](https://vuejs.org/guide/quick-start.html) instructions for doing so.
+
+## What is a Vue Component? Why should I care? <a name="what"></a>
+The Vue.js framework is a Javascript component-based programming model for the development of User Interfaces (UI), and is commonly used for Web Applications. It builds on the syntax of HTML, allowing the user to "declaratively" describe what they want the code to do. This allows for a very reactive app, one that responds efficiently to state changes to update the DOM. What this means is, less work needs to be done by the programmer for the HTML to update. The API takes care of a lot of it!
+
+Components are how it's all done. Every .vue file is a Vue Component! In these files, you can condense the 3 key aspects of a webpage, the HTML, code-behind, and CSS styling. This centralization and user-friendliness has made Vue renowned for its simplicity.
+
+**Am I quoting too much from official documentation? Should I put my own spin on it or cut out general info about the Vue.js framework?**
```
Firebrandt, line 23: Honestly, it's fine as-is I think.
On putting your own spin on it -- this isn't an essay, we're trying to teach people how to use Vue, and the official docs are really good at that so emulating them makes sense.

If you want to keep your guide laser focused (and you probably should) -- removing the general info would be sensible. We assume after all the user has a broad sense of what Vue is and how to set it up.

Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+<style scoped>
+button {
+  font-weight: bold;
+}
+</style>
+```
+The top is the Javascript code-behind, the middle is the HTML, and the bottom is the CSS style. Nice and centralized. Thankfully, the order doesn't matter. However, as the script is the most complicated, we'll talk about it last.
+
+**Should I explain what the example code above means? Like, what import {ref} does? Or maybe I should give a different example, one similar to what we worked on in our CSC301 project? I worry it might be too long**
+
```
Firebrandt, line 51: Example code is a good idea.

It might be sensible to split it up, though. Start by explaining just the HTML template, and showing just that part of the example.

Then the style part in the style section. Then the script part in the script section, then put it all together at the end.
The reason I ask for this is because I think it'd make sense to use as little working memory as possible -- we don't want people to have to keep the script example in memory until they get to the bottom of the page.

On example code -- assume the user knows what an import statement does -- do explain vue features that apply to components like state and stuff in your actual draft. Try to avoid going into too much detail because we could always just #read the docs if we need that.

Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+### HTML Template
+Ever seen the HTML of a webpage? That and Vue's HTML template are almost identical. Just remember, the root container must be labelled as ```<template>```, and it can only have one container inside that. Inside *that* sub-container though, anything is possible! On top of things like your standard ```<div>```, you can also use **other** Vue Components that you've imported in the script section! More on that in a bit.
+
+Here, I'm going to *only* show how Vue HTML components have if-else visibility features, and the ability to insert components through built-in for loops. Am I missing anything important to do with HTML in Vue?
+
```
Firebrandt, line 56: Not as far as I know -- we basically just use HTML to define what the component renders to and to add a little logic like props/callbacks etc. But I'm not a Vue expert so grain of salt.

Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+### CSS
+The CSS styling in Vue works as it does anywhere else, but with one key difference. See that ```scoped``` attribute in the ```<style>``` tag? When the tag has that attribute, the CSS can only apply to the component it's present in. Removing it however, allows for usage anywhere else in the program. Just...be careful about CSS conflicts and shadowing.
+
+### Script
+This is the meat of Vue. This is where you essentially define the webpage as an Object-Oriented-Programming Class! It can have it's own methods, attributes, as well as import scripts and other components for use! It's a very dense part of learning Vue, but here are the basics.
+
+Here I'm going to explain how to mount the app, basically talking about props, the default{} variables, computed methods, import statements, etc. It's going to be a rather dense section, as I have to talk about class variables and how mounted() works at runtime. Should I go into asynchronous methods/components?
+
```
Firebrandt, line 64: I think keep it as a quick primer for now -- this looks good -- but skip the async stuff. I think your goal is a basic guide, per our conversations earlier?

Topics/Tech_Stacks/Vue_Component_Setup_and_Tips
```diff
@@ -0,0 +1,73 @@
...
+## Things to watch out for! <a name="other"></a>
+Vue's reactivity and unique way of handling things can prove to be a double-edged sword if not handled properly. Here are some general tips:
+- If you want to display a document variable inside a component, and that component has an attribute that computes that same variable, you may cause Vue to enter an infinite update loop, where the computing of the variable will cause the displaying of the variable to cause the program to see the component as updated, so it recalculates the variable and redisplays it, and so on. This can crash the page or prevent it from displaying.
+- Uh what else?
```
Firebrandt, line 68: How do we solve this? It'd be handy to know. 
(I believe the solution is to make a child component have a prop).

Fight-Fighter:
Thanks for the tips, I guess I did go a bit overboard. And look at that, I also missed adding a link to the file in the Tech_Stacks.md file as well! Lemme just fix that.

Fight-Fighter:
Alright, I did what you suggested and expanded on the solution for that tip at the end, skipped the async stuff, and added in more example code! Last question though, how does it all look? I'm concerned with the ordering of the HTML and Script sections, since it sort of requires the reader shuttling back and forth between them to gain the full picture. Have I made each section standalone enough, or should I cut down or expand upon either?

Firebrandt:
> Alright, I did what you suggested and expanded on the solution for that tip at the end, skipped the async stuff, and added in more example code! Last question though, how does it all look? I'm concerned with the ordering of the HTML and Script sections, since it sort of requires the reader shuttling back and forth between them to gain the full picture. Have I made each section standalone enough, or should I cut down or expand upon either?

Yay, good job integrating my feedback.

I think the ordering of the HTML and script sections is okay. I mean, there are gonna be tradeoffs - do we make people remember more stuff, or do we keep stuff isolated? And I think keeping stuff isolated to only what's relevant makes the most sense. (It also doesn't require that much shuttling).

I think sections are sufficiently standalone and don't really require additional edits. I like it! 

I can't comment on how it looks because **you haven't made it an .md file.** For it to display properly, please do that lol.

Fight-Fighter:
Oops, thanks, now its an md file!

Firebrandt (approved):
All looks good! The layout is nice now that it's an .md

Fight-Fighter:
Thanks, time to merge!

PR #379, opened 2023-12-01
Introduction to Django Templates

darlynnguyen:
- Introduced Django templates and their uses
- Demonstrated Django Template Language's basic syntax and functionalities through examples and walkthroughs


Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+    └── templates/
+        └── index.html
+    └── ...
+```
+
+## Context
+Context is a dictionary of dynamic data that is passed to a template upon rendering. Later we will see a concrete example on how to do this in a function-based view.
+
+## Basic Syntax
+Django templates for webpages involve both static HTML tags, and DTL specific [syntax](https://docs.djangoproject.com/en/4.2/topics/templates/#syntax) including variables, tags, and filters.
```
qwet1235, line 47: Small typo: "DTL specific" -> "DTL-specific"?

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Variables
+**Variable Substitution** is done by wrapping the variable name (passed in from the context) with double curly braces: 
+```
+<li>{{ menu_item }}</li>
+```
+So this would render as a list item whose text is the value of `menu_item`. If this variable doesn't exist in the given context, then its value is `None` by default, and it would simply render an empty string. However, this can be [configured](https://docs.djangoproject.com/en/1.11/ref/templates/api/#how-invalid-variables-are-handled).
+
+We can also pass in lists and dictionaries. Using a `.` performs a lookup to access by index/attribute:
```
qwet1235, line 56: I think the second sentence is a bit confusing. Access what by index/attribute?

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+```
+<li>{{ menu_items.0.name }}</li>
+```
+
+Assuming menu_items is a list of dictionaries, this would display the first dictionary's name value.
```
qwet1235, line 61: I think it would be nice to put "menu_items" in a code block to maintain consistency

qwet1235: Same for "name" maybe?

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Built-in Tags
+DTL's built-in tags is a convenient way to allow for conditional statements, loops, loading static files, and [more](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/). Tags are characterized by the `%` symbol inside curly braces.
```
qwet1235, line 64: "is a convenient" -> "are a convenient"

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+
+#### Conditional Statements
+The syntax and [boolean operators](https://docs.djangoproject.com/en/4.2/ref/templates/builtins/#boolean-operators) for conditionals inside the tags are similar to Python, except it must end with an `{% endif %}` tag.:
+```
+{% if menu_item.is_special %}
+    <li class="special">{{ menu_item.name}}</li>
+{% else %}
+    <li class="special">{{ menu_item.name}}</li>
```
qwet1235, line 72: Were the code blocks inside "if" and "else" meant to be the same?

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+```
+With this setup, we have to configure `settings.py`:
+```
+# Static files (CSS, JavaScript, Images)
+# https://docs.djangoproject.com/en/4.2/howto/static-files/
+STATIC_URL = 'static/'
+STATICFILES_DIRS = [ BASE_DIR / 'static', ]
+```
+
+`STATICFILES_DIRS = [ BASE_DIR / 'static', ]` tells Django to look in the `static` folder in the root directory. However, there are times where you may want to have `static` directories in each app directory, which can be configured accordingly.
```
qwet1235, line 109: I think it would be helpful to include a link on how to configure this

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+
+In the template, we must load the static files with the `{% load static %}` tag, and include a specific file using the `{% static %}` tag. Using the above file structure, we can include the `.css` file:
+```
+{% load static %}
+<link rel="stylesheet" href="{% static 'css/styles.css' %}">
+```
+Note that the `{% load static %}` tag must be used before any `{% static %}` tag, so it is generally placed at the top of the template.
+
+### Built-in Filters
```
qwet1235, line 118: I think having this subsection before "Including CSS Files" would be better since you were talking about variables earlier

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
```diff
@@ -0,0 +1,193 @@
...
+
+The following puts the length of `list` in a paragraph:
+```
+<p>{{ list | length }}</p>
+```
+
+## Putting it together
+Revisiting our initial goal, we want to make WcDonald's webpage more configurable and dynamic. First, we create a template `menu.html` inside the app's `templates` folder. We want to allow changes to menu items, and control whether or not they are a special item.
+
+In our body, we can use a loop to display an abritrary number of menu items, and conditionals to decide whether they are a special or not. Additionally, to change the visibility of special items, we can include our `.css` files in the head:
```
qwet1235, line 129: Remove comma after "number of menu items"?

qwet1235 (changes requested):
There are some minor issues to fix, but the page is very good overall.  I really liked your diagrams and examples; they made the text easy to follow.

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
darlynnguyen: The logical flow of introducing the syntax follows the way I listed them in the introductory paragraph "Basic Syntax". So it makes sense to me to leave it after the "Built-in Tags" section.

Topics/Tech_Stacks/Django_Templates.md
```diff
@@ -86,37 +86,29 @@ We can loop through lists, dictionaries, and its nested variants with the famili
...
-With this setup, we have to configure `settings.py`:
-```
-# Static files (CSS, JavaScript, Images)
-# https://docs.djangoproject.com/en/4.2/howto/static-files/
-STATIC_URL = 'static/'
-STATICFILES_DIRS = [ BASE_DIR / 'static', ]
-```
-
-`STATICFILES_DIRS = [ BASE_DIR / 'static', ]` tells Django to look in the `static` folder in the root directory. However, there are times where you may want to have `static` directories in each app directory, which can be configured accordingly.
+With the above setup, we have do not have to configure `settings.py`. However, you may want a `static` folder in other locations, such as the `root` directory. In this case, you must configure `settings.py` [accordingly](https://docs.djangoproject.com/en/4.2/howto/static-files/#).
```
qwet1235, line 101: Small typo

qwet1235 (changes requested):

Topics/Tech_Stacks/Introduction_to_Django_Templates.md
qwet1235: I see, thanks for clarifying

PR #381, opened 2023-12-02
Added new content to Django rest framework topic and added the link to the readme file

VictorYao17:


Topics/Tech_Stacks/Django_Rest.md
```diff
@@ -46,13 +49,121 @@ Django Rest Framework is a popular library for building RESTful APIs with Django
...
+    ]
+}
+```
+This allows all your API endpoints to be used by authenticated and unauthenticated requests. However, this should only be used when testing and proper authentication and permissions should be added once the application is ready for production. "REST_FRAMEWORK" is also where you will place any other default configurations such as default authentication, renderers, pagination, etc.
 
- ### How it Works
- 
+
+### Serializers
+Since Django Rest Framework sends and receives data in JSON format, we need a way to convert Django Model data into a form that can be stored and transported in JSON format. This is exactly what serializers do. Serializers convert model data into JSON format and data received in JSON format back into model instances. Additionally, serializers also do data validation. To use serializers it is best practice to create a new file in your app called "serializers.py" and write all your serializers in that file.
```
roryhofland, line 72: Maybe add a small example of model data being serialized into JSON and vice versa.

Topics/Tech_Stacks/Django_Rest.md
```diff
@@ -46,13 +49,121 @@ Django Rest Framework is a popular library for building RESTful APIs with Django
...
+ ```
+INSTALLED_APPS = [
+    ..., 
+    'rest_framework',
+]
+ ```
  
+ While in the settings.py file, you should also add:
+```
+REST_FRAMEWORK = {
```
roryhofland, line 62: You could add an external link aroudn this section that points towards more configuration options that go in the settings.py file.

VictorYao17: Thanks for the suggestion, I will add that

Topics/Tech_Stacks/Django_Rest.md
VictorYao17: I have just added an example, let me know what you think.

roryhofland: Looks good!

PR #397, opened 2023-12-03
Added overview of Caddy

Connor22:
- Added a page on Caddy to the wiki
- added link to new page in the `Topics/Development_Process.md` index page
- modified `Topics/Development_Process.md` by combining similar links in index into a single group.

### DIFF
--- a/Topics/Development_Process.md
+++ b/Topics/Development_Process.md
@@ -26,13 +26,19 @@
 
 ### [Getting Started With WSL 2](./Development_Process/WSL.md)
 
-## Nginx
+## Accessing your Project
 
-### [Introduction to Nginx](./Development_Process/Nginx.md)
+### Nginx
 
-## DuckDNS
+#### [Introduction to Nginx](./Development_Process/Nginx.md)
 
-### [Overcoming Dynamic IPs with DuckDNS](./Development_Process/Duckdns.md)
+### Caddy
+
+#### [Getting Started with Caddy](./Development_Process/Caddy.md)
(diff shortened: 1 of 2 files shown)


gtian239 (changes requested):
Good work with the article! It gives a comprehensive overview of Caddy. I also like the reorganization of the related topics in `Topics/Development_Process.md`.

A couple comments/questions for `Topics/Development_Process/Caddy.md`:

-The `Installation` section provides enough to figure out how to install Caddy, but I think you should include a reference link to the installation documentation on their [website](https://caddyserver.com/docs/install) or [GitHub repository](https://github.com/caddyserver/caddy#install) for direct guidance.
- On line 18-20, since you have explained `--to :9000`, could you also clarify the meaning of `--from :80`? I'm not familiar with Caddy, so I'm not sure if this part is significant enough to need to expand on.
- On line 23, there is an example heading, but no example.
- Could you provide a brief overview of Caddyfile before you start explaining how to use it? For example, moving parts of line 52 to the beginning or before the `Running with a Caddyfile` section so people who have never used Caddy can easily understand the purpose of the commands in this section.
- Could you expand on which "options are required" in line 127? The previous references to options are "global options", so I assumed these are the options you are referring to, but the example below this line has brackets around the global options. Could you please clarify which options you are referring to, and if the example is related? 

Otherwise, great work! For someone who is not familiar with Caddy, the content is mostly clear and straightforward and would serve as a good guide for Caddy.

PR #403, opened 2024-02-23
Create Machine Learning Model Deployment.md

matsevytyi:
Wrote a report about ML deployment as A2 for CSC301

It was a new but crucial and interesting topic for me


stalight:
Generally it's an informative page, but it contains a lot of words non-ML student or people who might want to get into ML may have trouble with because they may have 0 clue what they are, so it's important to add hyperlink and/or explain what they are to the audience.

matsevytyi:
Firstly, I can't imagine a person deploying ML without any basic understanding of this field.

Nevertheless, I have provided links to possible resources (courses and books), that don't require user to have some skills in the field, where they can catch the basic understanding as well.

PR #476, opened 2024-03-17
Creating 3D objects with threejs

Lenzs:


jjackzhu:
Hi Bill! One suggestion would be to add an example picture of what the end result should look like, as well as some images that show off what other have done using three.js!

PR #480, opened 2024-03-17
Ubuntu Server 20.04 guide

RuilinP:
The original wiki has included several posts covering the setup of web app on a cloud server, while this one is for the purpose of showing the possibility of home server using a popular Linux distro called "Ubuntu Server Edition". The guide includes a brief introduction of motivation, the technical requirements, information of the OS, Internet connection, package installation, SSH, an example of hosting a web app using the Nginx, and further reading consisting related topics of next steps or further details. The detailed instructions is believed to be friendly towards non-terminal users.

> **Changes**
- Added a folder named "Ubuntu_server_edition" under Topics>Development_Process. Within the folder is the "ubuntu.md" which is the document and a sub folder called "Assets" with the screenshots used by the document.
- The "ubuntu.md" include the following sessions: Introduction, Requirements, About Ubuntu Server LTS 20.04, bout Ubuntu Server LTS 20.04, Connection to Internet, Installation of necessary tools using apt, SSH, Web Server Example, Further reading.

> **Justification**
Like included in the document, usage of cloud hosting is not cheap and neither can most of us make full use of their functionalities. Therefore, home server is considered to be a better option. And given that all bahen CDF labs run ubuntu, the ubuntu server would be the OS choice we can get our hands on the quickest.
image


benxu3:
overall an interesting and well documented read, I definitely learned something new and I think this would be a valuable add to the repository

Issue #421, opened 2024-03-13
Tech Stack: Learning JSON-RPC


WilliamJiH:

roxanehe:
I am waiting to review :)

PR #481, opened 2024-03-17
feat: Learning JSON RPC with ETH #421

WilliamJiH:


roxanehe:
Looks good to me! learned about this through real life examples.

Issue #483, opened 2024-03-17
Tech Stacks: Selenium UI - Automating a Google Form Tutorial


LanceUy:
This tutorial aims to build upon the current Selenium page (under Tech_Stacks) in the wiki by giving a more comprehensive example on using Selenium UI to automate user actions. In this example, I will be automating a Google Form using Selenium. In today's software development landscape, rigorous testing is imperative before delivering code to customers. Whether it's through API testing or UI testing, major tech companies prioritize comprehensive testing to ensure product quality. Roles like Software Engineer in Test (SET), which I've had the privilege to work in during my PEY, heavily rely on expertise in this area.

PR #489, opened 2024-03-17
Tech Stack - Updated selenium.md with full simple tutorial

LanceUy:
Pull request for update on selenium.md #483. Updated with full tutorial example that goes through the basics of Selenium UI Automation as well as added a new folder containing image for examples.


Topics/Tech_Stacks/selenium.md
```diff
@@ -124,6 +124,149 @@ driver.quit()
...
+    textField.click()
+    textField.send_keys("Hello")
+    ```
+3.  **Selenium: Selecting Multiple Choice**
+    
+    Let's say we wanted to select 'Option 2'. Following similar steps as above we will first locate the css selectors. Inspecting the multiple choice option for 'Option 2' we see the following:
+
+    ![alt text](SeleniumImages/multipleChoice.png)
+
+    We see there is a unique id field ```id = "i12"```which we can use for our css selector. We want to ensure that our css selectors are unique so that we choose only the item we want to select. As such below is the example code as how we could go about doing this.
```
harryye930, line 204: I would suggest to add a bit more information on how to identify/choose what field to include in the CSS selector, as some of the fields can be dynamically generated as part of the JavaScript code.  I.e., The third option under a <div> with classname="something"

harryye930 (commented):
I like your tutorial to automate a common web application, with such interactive guide and step-by-step process!

Maroosh-Gillani (commented):
Great job! I am very impressed with your contribution.
I think adding the following changes can make your contribution even better:
- I can see that you have added multiple code fragments, but many of them are not explained explicitly. Adding an explanation will increase clarity.
- Consider adding a conclusion section that combines all the code sections into one.
- From my understanding, Selenium is very prone to errors, so this contribution could maybe benefit from adding a small debug section.

Maroosh-Gillani (approved):
Good job, looks good to me!

PR #495, opened 2024-03-17
Added two demo videos to design a page and link between pages in the app using Figma

yumnarefai:
Embedded my two demo videos on using Figma to design and navigate an app. I provided a clear example showcasing how to use Figma to design a high fidelity prototype for an app and link pages across the application. This is done at a high level so that developers attain a clear understanding and plan of the process to begin coding the app.


helias22 (approved):
Thank you for your thorough demo. It was a great idea to include some visual and video demonstrations instead of just text explanations. With the use of visuals, I'm pretty sure anyone like me, with no background in figma creation, will be able to start their own prototype and link the pages.

As a next-step suggestion, while you were explaining the linking process, I realized that the footer exists on all pages. I was wondering, do you have to manually link it on every page, or is there a way to have it applied universally?

This would, of course, not be considered a beginner functionality, and I wasn't expecting to see it in this demo.

PR #512, opened 2024-03-17
TTS - Google Cloud API

TianTan2024:
added google cloud information and conclusion


Everett02 (changes requested):
Tian, you did a great job on describing the Google Cloud TTS, together with introducing the way to set up environment and get the api. It will be better if you can add some simple usecase samples, which will make it more comprehensive.

TianTan2024:
> Tian, you did a great job on describing the Google Cloud TTS, together with introducing the way to set up environment and get the api. It will be better if you can add some simple usecase samples, which will make it more comprehensive.

greate idea, I will do it now. Thank you for review!

Everett02 (approved):
It looks good now. Thank you for your changes!

PR #516, opened 2024-03-17
Introduction to sublime

ghost:
I created a post about the introduction to Sublime. I also explained how to install the software tool with specific guides and key features of the tool.


carolinexuxy:
Nice work. The overall document is clean and easy to understand. The gifs are really nice visuals that help the readers to understand the process. 

A couple of things I think it's worth to consider: 
1. It's worth to mention "why sublime text" and highlights a couple properties of sublime text that makes it stands out from other text editors. (e,g, easy to user, simple UI etc.) This will brings a lot of insights when someone is considering which which editors to use between sublime text and some other editors. 

2. The resources links to some really helpful sites, but since the links are long, it may be more intuitive to reorganize them or/and even put them into hyperlink with a title. This will help the reader to navigate through the resources easier and faster.

carolinexuxy (changes requested):
Nice work. The overall document is clean and easy to understand. The gifs are really nice visuals that help the readers to understand the process.

A couple of things I think it's worth to consider:

It's worth to mention "why sublime text" and highlights a couple properties of sublime text that makes it stands out from other text editors. (e,g, easy to user, simple UI etc.) This will brings a lot of insights when someone is considering which which editors to use between sublime text and some other editors.

The resources links to some really helpful sites, but since the links are long, it may be more intuitive to reorganize them or/and even put them into hyperlink with a title. This will help the reader to navigate through the resources easier and faster.

carolinexuxy (commented):
Looks good now, please merge in after resolving the conflict

carolinexuxy (approved):

Issue #419, opened 2024-03-11
System Design: A Guide to OpenMP


leowrites:
Understanding and designing parallel systems is crucial in today's computing landscape, where scalability and efficiency are paramount. This guide aims to delve into parallel systems, offering insights into their architecture, implementation, and optimization.

maxkmy:
I will be working on this with Leo.

justinwli930:
I will be working on this as well

PR #522, opened 2024-03-17
add: openMP introduction

maxkmy:
- add parallel system introduction
- add prerequisites
- add shared memory
#419


justinwli930 (commented):
Looks good Max! One suggestion would be that some students might not already be familiar with paradigms of concurrency, effectively overshadowing the importance of parallel programming and this guide overall. I would suggest maybe adding a brief section on concurrency, or adding a few resources that students can go and read about (this is probably a better option since concurrency is probably not a simple topic to learn about for the first time).

justinwli930 (approved):
lgtm! Thanks max

PR #526, opened 2024-03-17
add openmp overview

justinwli930:
- add openmp overview
- add openmp programming model
- add very simple examples

#419


maxkmy (changes requested):
Looks good. For the calculating the sum of an array in parallel, would it be possible to add a sequential version? I think it makes understanding parallel code in OpenMP is not so different from sequential code.

justinwli930:
> Looks good. For the calculating the sum of an array in parallel, would it be possible to add a sequential version? I think it makes understanding parallel code in OpenMP is not so different from sequential code.

Good point! Will add it.

maxkmy (commented):
Looks good. Thank you for making this change!

leowrites:
Looks good. I like the examples but could you explain a bit about the clauses that you are using (specifically what `reduction` and `for` are doing?)

justinwli930:
> Looks good. I like the examples but could you explain a bit about the clauses that you are using (specifically what `reduction` and `for` are doing?)

Good point. Will do!

maxkmy (approved):
Looks good to me.

leowrites (approved):

PR #528, opened 2024-03-17
Introduction to Pandas

ryanMMM:
- Added Introduction to Pandas file which gives an intro into data reading, manipulation, analysis, and visualization
- Added link to the file in tech stack list markdown file


epicmine99 (changes requested):
I have to say, a great introduction to Pandas Ryan! I was thoroughly engaged the whole time as you covered the basics of Pandas. Below is some feedback I think you'll appreciate to improve on your getting started guide!

### Pandas vs Other data processing libraries (like Numpy or R)
Pandas is pretty unique in the way it handles data and differs from other related libraries making it hard for Numpy and R veterans to transition into. From what I can remember  Pandas and Numpy have different data storage paradigms and pandas handle operations column-wise, whereas Numpy and R handle operations row-wise. small things like that could save beginners a lot of headaches and I am sure that there are more differences than that.

edit: after doing some research my example is actually wrong according to this source: https://aman.ai/primers/pandas/#pandas-is-column-major

Regardless there are important differences between Pandas and other data processesing libraries that might be important to point out, or maybe even explaining what column major or row major is to beginners!

### SQL
Given Pandas is a great library for processing tables of data, programmers often need to export the data into databases using SQL. Given its popular use case, you can expand on the SQL aspect and cover cases where you would use SQL or how Pandas interacts with SQL libraries. 

### Small correction
In the first sentence of the conclusion you used the term 'wiki' when it would make more sense in the context of the sentence to use a term like 'page'. You could also use a new line before your references to distinguish them from your conclusion

epicmine99 (approved):
Great changes Ryan, thank you!

PR #535, opened 2024-03-17
Create logging_to_mysql.md

hantien2002:


Tahasp:
Hey, would you be up to reviewing each other’s work?

hantien2002:
@Tahasp Hi, sorry I have already found teammate to review my work.

OliviaYZhou:
Hi Tien, the PR looks good overall, I like the visuals and images you included, especially the video. 
One thing you could add is you can explain your logs table in Step 2. For instance, the user might not know what a level is or why it needs to be added. You have a pretty good explanation for users table in part 3, maybe you can do something similar for logs table.
On a similar note, since you are adding a users table, but not all logging requires a user, you can add a section explaining a use case where someone would need a users table, and explain that this tutorial will be most helpful in similar cases.

OliviaYZhou (changes requested):
Sorry I commented in wrong place. This is the PR review.

Hi Tien, the PR looks good overall, I like the visuals and images you included, especially the video.
One thing you could add is you can explain your logs table in Step 2. For instance, the user might not know what a level is or why it needs to be added. You have a pretty good explanation for users table in part 3, maybe you can do something similar for logs table.
On a similar note, since you are adding a users table, but not all logging requires a user, you can add a section explaining a use case where someone would need a users table, and explain that this tutorial will be most helpful in similar cases.

Topics/Software_Engineering/logging_to_mysql.md
```diff
@@ -0,0 +1,196 @@
...
+
+```sql
+CREATE TABLE logs (
+    id INT AUTO_INCREMENT PRIMARY KEY,
+    timestamp TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+    level VARCHAR(10),
+    message TEXT
+);
+```
+
```
OliviaYZhou, line 33: You can add an explanation for the columns here

Topics/Software_Engineering/logging_to_mysql.md
```diff
@@ -0,0 +1,196 @@
...
+## Usage Example with the `users` Table
+
+In this example, we have a `users` table in our MySQL database designed to store user information such as IDs, names, and email addresses. 
+The table structure is as follows:
+
+- **id** (INT): A unique identifier for each user, auto-incremented.
+- **name** (VARCHAR): The name of the user.
+- **email** (VARCHAR): The user's email address, which is unique across all users.
+
```
OliviaYZhou, line 92: Maybe add a use case for an app that would require logging and also has a users table around here

OliviaYZhou (changes requested):

hantien2002:
@OliviaYZhou Hi Olivia, thank you for your review and I have made changes based on your suggestion.

OliviaYZhou:
Hi Tien, I've seen the changes and they look great. The explanations make the tutorial clearer. This will be a great guide for anyone who wants to create logging using mysql.

PR #543, opened 2024-03-17
add limitations

leowrites:
- added Amdahl's Law
- added overhead due to parallelization
- added false sharing and its impact on performance


maxkmy (changes requested):
Looks great for the most part. I think it would be nice to briefly give an intuitive description of what a cache line is. Otherwise, there may be some knowledge gaps when discussing false sharing.

maxkmy (approved):
Looks good. Thanks for adding this!

PR #544, opened 2024-03-17
Update p5.md - Included the introduction, usage and advantages of p5.js

sinemsoylemezoglu:
Included my portion which is the introduction, usage and advantages of p5.js; described the basic idea of p5.js and why it can be useful for the users to include this library in their project(s).


Tahasp:
Hey, would you be up to reviewing each other’s work?

nat0cat (changes requested):
Overall, this looks fine, however some points in Usage, and Advantages are redundant, and it may be more effective to combine them.

PR #565, opened 2024-03-18
Architecting a System to Handle 100k+ Requests

benxu3:
**Motivation**
When we build a project, we are often concerned with the product; what features does the user want? Once we've identified the features to build, we come up with a specification and implement our solution accordingly. Maybe we test our code to ensure functionality and correctness before then shipping our project.

But what happens after this? If we have have tens or hundreds of thousands of users each making lots of requests, how do we know that our system can handle the load? I hope to partly resolve this question in my contribution by providing a high-level overview of designing such a system.

**Topics**
- Vertical & Horizontal Scaling
- Load Balancer
- Database Scaling
- Caching
- CDN


Topics/System_Design/assets/load-balancer.png
RuilinP, line 1: I think there could be a easier to understand pic than that, for eg: https://limeproxies.netlify.app/blog/how-to-balance-traffic-without-server-getting-crashed
image

Topics/System_Design/Handling_100K_Requests.md
```diff
@@ -0,0 +1,52 @@
...
+## Caching
+![Redis Cache](./assets/redis-cache.png)
+*Image from Backendless*
+
+Another way we can reduce the load on our databases is by introducing a cache layer using something like [Redis](https://redis.io/). When a user makes a request for some data from the database, we store that information in the cache layer with some expiration time *t*. Any future requests made for that information within time period *t* are actually hitting the cache rather than the database. This reduces the workload of our database by handling some of it with the cache layer, and we can scale our cache layer as well by using distributed caching and storing data across multiple machines. 
+
+
+
+## Content Delivery Networks (CDNs) 
+CDN servers keep a cache of our files (images, HTML/CSS, JS, ...) at data centers close in geographical proximity to users across the world. When user's make a request for data, it is retrieved from the server closest to the user. This again offloads some of the requests from the main server and helps with reducing errors by increasing redundancy.
```
RuilinP, line 48: Since you have used picture to illustrate ideas for all previous sections, I think you probably want to use one for CDN as well. There are a wide range of uses of CDNs, from picture/manga websites to frameworks like bootstrap.
image

image

RuilinP:
Otherwise, generally easy to understand

RuilinP (commented):
Otherwise, generally easy to understand

Topics/System_Design/assets/load-balancer.png
benxu3: thanks for linking the image, I do agree it's much more explicit and made the change

Topics/System_Design/Handling_100K_Requests.md
benxu3: thanks I included another image as you suggested

PR #569, opened 2024-03-18
Create nano.md

Rakan1837:
Added instructions on editing, saving, and closing.


Tahasp:
Hey would you want to review each other's work?

Tahasp (commented):
Generally looks good but you need a bit more detail. I think example use cases and pictures would really elevate the understanding. Also some neat tips and tricks would be helpful!

PR #573, opened 2024-03-18
Added section on component customization

Eggathin:


Affixrevy:
I like the specific examples added here. It adds some needed detail onto the post and makes it a more comprehensive guide to people getting started with material ui. Good work!

PR #574, opened 2024-03-18
Update p5.md

nat0cat:
Added basic components of p5.js as well as a description of most commonly used functions and example usages.


kimmy190 (commented):
The core elements of p5.js are well organized into sub-topics, providing beginner-friendly insights. I like the inclusion of code snippets alongside each function, enhancing the usability for readers as it is easier for them to directly test them out.

